PR #1, opened 2020-01-10
[tests] upload coverage

wgtdkp:
This PR uploads coverage to https://codecov.io.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,11 @@
+coverage:
+  status:
+    project: off
+    patch: off
+
+ignore:
+  - "tests/*"
+  - "third_party/*"
+
+comment:
+  layout: "diff, flags, files"
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,8 +4,8 @@ coverage:
     patch: off
 
 ignore:
-  - "tests/*"
-  - "third_party/*"
+  - "tests/**/*"
+  - "third_party/**/*"
 
 comment:
   layout: "diff, flags, files"
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -26,7 +26,7 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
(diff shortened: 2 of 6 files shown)


.codecov.yml
```diff
@@ -0,0 +1,11 @@
+coverage:
+  status:
+    project: off
+    patch: off
+
+ignore:
+  - "tests/*"
+  - "third_party/*"
```
bukepo: should use third_party/**/*

tests/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
-if (CMAKE_BUILD_TYPE AND OT_COMMISSIONER_COVERAGE AND CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
+if (CMAKE_BUILD_TYPE AND OT_COMM_COVERAGE AND CMAKE_CXX_COMPILER_ID MATCHES "GNU|Clang")
```
bukepo: OT_COVERAGE?

wgtdkp: Should we use `OT_COMM_` as prefix of all CMake options in OT Commissioner? There will be name conflict if a project includes both `openthread` and `OT Commissioner` as a library.

bukepo: I assume these are common flags. Just like DEBUG. User don't have to set OT_DEBUG and OT_COMM_DEBUG. Thoughts?

wgtdkp: No, this has an assumption on libraries that they using the same options prefix. I believe it is not the usual case for CMake users. It is always for CMake user to explicitly config each dependency independently. Agree?

jwhui:
@wgtdkp, the codecov.io upload does not seem to be happening.

wgtdkp:
@jwhui It seems the travis build is fetching wrong commit https://travis-ci.org/openthread/ot-commissioner/builds/652305327?utm_medium=notification&utm_source=github_status. Investigating.

wgtdkp:
This seems to be a [bug of travis](https://travis-ci.community/t/build-checks-out-wrong-pull-request-79-instead-of-7/6118). Will close this PR and create a new one.

PR #2, opened 2020-01-15
[tests] remove 1.2 CCM integration tests before dependencies becoming public

wgtdkp:
This PR removes 1.2 CCM integration tests.


tests/integration/bootstrap.sh
```diff
@@ -48,47 +47,49 @@ setup_otbr() {
...
     set -e
-    git clone ${OPENTHREAD_CCM_REPO} ${OPENTHREAD_CCM} --branch ${OPENTHREAD_CCM_BRANCH}
+    git clone ${OPENTHREAD_REPO} ${OPENTHREAD} --branch ${OPENTHREAD_BRANCH}
 
-    cd ${OPENTHREAD_CCM}
+    cd ${OPENTHREAD}
 
-    #./script/bootstrap
-    rm -rf build output
+    #./script/bootstrap || true
```
librasungirl: remove if useless?

tests/integration/bootstrap.sh
```diff
@@ -48,47 +47,49 @@ setup_otbr() {
...
+            BORDER_ROUTER=1 \
+            SERVICE=1 \
+            DHCP6_CLIENT=1 \
+            DHCP6_SERVER=1 \
+            JOINER=1 \
+            COMMISSIONER=1 \
+            MAC_FILTER=1 \
+            REFERENCE_DEVICE=1 \
+            THREAD_VERSION=1.2 \
+            REFERENCE_DEVICE=1 \
```
librasungirl: nit: dup with Line # 71

tests/integration/bootstrap.sh
```diff
@@ -48,47 +47,49 @@ setup_otbr() {
 
 setup_openthread() {
     set -e
-    git clone ${OPENTHREAD_CCM_REPO} ${OPENTHREAD_CCM} --branch ${OPENTHREAD_CCM_BRANCH}
+    git clone ${OPENTHREAD_REPO} ${OPENTHREAD} --branch ${OPENTHREAD_BRANCH}
```
simonlingoogle: What if OpenThread updates cause ot-commissioner tests to fail ? Should we use a fixed stable version of openthread? Thoughts? @librasungirl

tests/integration/common.sh
```diff
@@ -27,13 +27,24 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
+## This controls with which test suite to run.
+## Set environment variable `OT_COMM_TEST_SUITE`
+## to 1.1 or 1.2 to override this.
+readonly TEST_SUITE=${OT_COMM_TEST_SUITE:="1.2"}
```
simonlingoogle: Since 1.2 is not released, should we use 1.1 as default? @librasungirl

tests/integration/bootstrap.sh
librasungirl: On one hand, currently we don't have a regular stable release process; on the other hand, there is no test against external commissioner in OT's repo, however we may want to ensure the stability of master branch if possible. So it might also be an early warning to ensure the quality of OT if there is fails.
Thoughts?

wgtdkp: @simonlingoogle @librasungirl Agree that integration tests here can reveal openthread bugs related to external commissioning. We may leave this to future enhancement to get an ideal solution Here let's get this blocking PR merged first. Thoughts?

tests/integration/bootstrap.sh
```diff
@@ -48,47 +47,47 @@ setup_otbr() {
...
-        BORDER_AGENT=1 \
-        UDP_FORWARD=1 \
-        DEBUG=1
-
-    cp output/x86_64-unknown-linux-gnu/bin/ot-cli-ftd ${CCM_CLI}
-    cp output/x86_64-unknown-linux-gnu/bin/ot-ncp-ftd ${CCM_NCP}
-
-    executable_or_die "${CCM_CLI}"
-    executable_or_die "${CCM_NCP}"
+    if [ "${TEST_SUITE}" = "1.2" ]; then
```
bukepo:
```suggestion
    if [ "${THREAD_VERSION}" = "1.2" ]; then
```

wgtdkp:
@bukepo @librasungirl @simonlingoogle removed 1.2 tests, please take a look again.

tests/integration/bootstrap.sh
librasungirl: agree

PR #3, opened 2020-01-15
[tools] fix Commissioner THCI dependency on pexpect

wgtdkp:
`make_local_commissioner` is for travis tests, `make_harness_commissioner` is for test harness. We don't want to add dependency to test harness.

### DIFF
--- a/tools/commissioner_thci/commissioner_impl.py
+++ b/tools/commissioner_thci/commissioner_impl.py
@@ -166,13 +166,13 @@ class OTCommissioner(ICommissioner):
         else:
             self._handler.terminate(force=True)
 
-    @classmethod
+    @staticmethod
     def make_local_commissioner(config, simulator):
         import pexpect
         handler = pexpect.spawn("/bin/bash")
         return OTCommissioner(config, handler, simulator)
 
-    @classmethod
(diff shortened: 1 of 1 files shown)


wgtdkp:
travis build depends on https://github.com/openthread/ot-commissioner/pull/2.

tools/commissioner_thci/commissioner_impl.py
```diff
@@ -169,11 +168,12 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 
     @classmethod
     def make_local_commissioner(config, simulator):
+        import pexpect
         handler = pexpect.spawn("/bin/bash")
         return OTCommissioner(config, handler, simulator)
 
     @classmethod
-    def make_remote_commissioner(config, serial_handler):
+    def make_harness_commissioner(config, serial_handler):
```
simonlingoogle: the first argument of an `classmethod` should be cls, which will be a reference to the class itself.
you should add the `cls` argument, or use `staticmethod` instead.

Same for `make_local_commissioner`

wgtdkp: `staticmethod` seems to be what I want.

PR #5, opened 2020-01-21
[cbor] fix potential "-Wnon-virtual-dtor" warning.

xiaom-GitHub:


src/library/cbor.hpp
```diff
@@ -51,6 +51,8 @@ namespace commissioner {
 class CborValue
 {
 public:
+    virtual ~CborValue() = default;
```
wgtdkp: Do we need to do the same for `cose.hpp:Object` class?

xiaom-GitHub: "-Wnon-virtual-dtor" reminds that if any virtual function already defined in the class, virtual destructor should be declared explicitly, to help avoid forgetting to release subclass when free the pointer which points to the base class. Looks like cose.hpp::Object is empty and no virtual functions defined in the cose.hpp::Sign1Message class, maybe no need to declare?

wgtdkp: I don't think so. Any base class in c++ should declare a `virtual destructor`.  Section 5.3.5 of [C++11 draft n3242](http://www.open-std.org/JTC1/SC22/WG21/docs/papers/2011/n3242.pdf) says:

>3. In the first alternative (delete object), if the static type of the object to be deleted is different from its dynamic type, the static type shall be a base class of the dynamic type of the object to be deleted and the static type shall have a virtual destructor or the behavior is undefined. In the second alternative (delete array) if the dynamic type of the object to be deleted differs from its static type, the behavior is undefined.

Take below snippet for example:

```c++
// clang++ -std=c++11 -Wall -g virtual_dtor.cc
#include <iostream>

class Foo {
public:
    ~Foo() {
        std::cout << "destructor of Foo" << std::endl;
    }
};

class Bar : public Foo {
public:
    ~Bar() {
        std::cout << "destructor of Bar" << std::endl;
    }
};

int main() {
    Foo *foo = new Bar();
    delete foo;
    return 0;
}
```

The destructor of `class Bar` will not be called.

wgtdkp: destructors in `cbor.hpp` and `cose.hpp` actually do nothing. So we don't need to worry about even if it will not be called, but this is wrong. To make things right and any eliminate potential error/warning, we need to make destructor of any base class `virtual`.

wgtdkp:
Depending on https://github.com/openthread/ot-commissioner/pull/20 to get travis builds passed.

jwhui:
# [Codecov](https://codecov.io/gh/openthread/ot-commissioner/pull/5?src=pr&el=h1) Report
> Merging [#5](https://codecov.io/gh/openthread/ot-commissioner/pull/5?src=pr&el=desc) into [master](https://codecov.io/gh/openthread/ot-commissioner/commit/af0c02bff2d590848dc0945dd81cfbf0034cca45&el=desc) will **increase** coverage by `0.00%`.
> The diff coverage is `n/a`.

```diff
@@           Coverage Diff           @@
##           master       #5   +/-   ##
=======================================
  Coverage   59.94%   59.94%           
=======================================
  Files          53       54    +1     
  Lines        5395     5400    +5     
=======================================
+ Hits         3234     3237    +3     
- Misses       2161     2163    +2     
```


| [Impacted Files](https://codecov.io/gh/openthread/ot-commissioner/pull/5?src=pr&el=tree) | Coverage Δ | |
|---|---|---|
| [src/library/cbor.hpp](https://codecov.io/gh/openthread/ot-commissioner/pull/5/diff?src=pr&el=tree#diff-c3JjL2xpYnJhcnkvY2Jvci5ocHA=) | `40.00% <0.00%> (-26.67%)` | :arrow_down: |
| [src/library/cose.hpp](https://codecov.io/gh/openthread/ot-commissioner/pull/5/diff?src=pr&el=tree#diff-c3JjL2xpYnJhcnkvY29zZS5ocHA=) | `100.00% <0.00%> (ø)` | |

Issue #7, opened 2020-01-28
Seg Fault when running on MacOS 10.14.6


ndrake:
The commissioner-cli is crashing when I try to run it on MacOS. 

I followed the build instructions described [here](https://github.com/openthread/ot-commissioner/blob/master/src/app/cli/README.md), and it built without issue.

I'm running it like this: `~/bin/bin/commissioner-cli ~/bin/bin/non-ccm-config.json`

Here is the crash report:

```
Process:               commissioner-cli [25583]
Path:                  /Users/USER/*/commissioner-cli
Identifier:            commissioner-cli
Version:               0
Code Type:             X86-64 (Native)
Parent Process:        zsh [9966]
Responsible:           commissioner-cli [25583]
User ID:               501

Date/Time:             2020-01-28 09:13:46.724 -0500
OS Version:            Mac OS X 10.14.6 (18G2022)
Report Version:        12
Bridge OS Version:     3.0 (14Y906)
Anonymous UUID:        6E71FCEE-C812-157B-4965-A2A9CDCA0C94

Sleep/Wake UUID:       BD18215C-33A4-4FCE-9AAD-230AFE0B8695

Time Awake Since Boot: 22000 seconds
Time Since Wake:       3900 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Signal:    Segmentation fault: 11
Termination Reason:    Namespace SIGNAL, Code 0xb
Terminating Process:   exc handler [25583]

VM Regions Near 0:
--> 
    __TEXT                 00000001062fd000-0000000106466000 [ 1444K] r-x/r-x SM=COW  /Users/USER/*

Thread 0 Crashed:: Dispatch queue: com.apple.main-thread
0   commissioner-cli              	0x000000010631209c ot::commissioner::CommissionerApp::~CommissionerApp() + 28
1   commissioner-cli              	0x000000010631a2f7 std::__1::__shared_ptr_pointer<ot::commissioner::CommissionerApp*, std::__1::default_delete<ot::commissioner::CommissionerApp>, std::__1::allocator<ot::commissioner::CommissionerApp> >::__on_zero_shared() + 23
[...]

PR #8, opened 2020-01-29
[app] fix destruction of commissioner app

wgtdkp:
This PR fixes the bug that `CommissionerApp` trying to stop the internal commissioner member.

This will result in **segment fault** if creating the `CommissionerApp` instance failed (`mCommissioner` is null), just as reported in issue https://github.com/openthread/ot-commissioner/issues/7.

### DIFF
--- a/src/app/commissioner_app.hpp
+++ b/src/app/commissioner_app.hpp
@@ -70,7 +70,7 @@ public:
     using Seconds      = std::chrono::seconds;
 
     static std::shared_ptr<CommissionerApp> Create(const std::string &aConfigFile);
-    ~CommissionerApp();
+    ~CommissionerApp() = default;
 
     // Discover Border Agent on link-local network.
     Error Discover();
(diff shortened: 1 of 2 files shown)


src/app/commissioner_app.hpp
```diff
@@ -70,7 +70,7 @@ class CommissionerApp
     using Seconds      = std::chrono::seconds;
 
     static std::shared_ptr<CommissionerApp> Create(const std::string &aConfigFile);
-    ~CommissionerApp();
+    ~CommissionerApp() = default;
```
Moandor-y: There won't be any problem if `mCommissioner` is not stopped?

If that's true, you can just remove it.
```suggestion
```

wgtdkp: @Moandor-y  `mCommissioner` will try to stop itself in its destructor.

keep `~CommissionerApp() = default;` to make it explicit that we are not forgetting the destructor. It serves a comment.

Issue #9, opened 2020-02-08
Add callback to indicate a joiner has been successfully commissioned.


wgtdkp:
As a library, OT commissioner needs to notify the application that a joiner is requesting and report the commissioning status.

jwhui:
Resolved by #11

PR #11, opened 2020-02-08
[api] add commissioning handler

wgtdkp:
This PR adds a new API to notify application that a joiner has been successfully commissioned:
```c++
using CommissioningHandler = std::function<void(const JoinerInfo &aJoinerInfo, Error aError)>;
virtual SetCommissionerHandler(CommissioningHandler aCommissioningHandler) = 0;
```

This addresses issue https://github.com/openthread/ot-commissioner/issues/9.

### DIFF
--- a/src/library/commissioner_impl.cpp
+++ b/src/library/commissioner_impl.cpp
@@ -159,6 +159,7 @@ CommissionerImpl::CommissionerImpl(struct event_base *aEventBase)
     , mPanIdConflictHandler(nullptr)
     , mEnergyReportHandler(nullptr)
     , mJoinerInfoRequester(nullptr)
+    , mCommissioningHandler(nullptr)
 {
     mBrClient.AddResource(mResourceUdpRx);
     mBrClient.AddResource(mResourceRlyRx);
@@ -268,6 +269,11 @@ const Config &CommissionerImpl::GetConfig() const
 void CommissionerImpl::SetJoinerInfoRequester(JoinerInfoRequester aJoinerInfoRequester)
 {
     mJoinerInfoRequester = aJoinerInfoRequester;
+}
+
+void CommissionerImpl::SetCommissioningHandler(CommissioningHandler aCommissioningHandler)
+{
+    mCommissioningHandler = aCommissioningHandler;
 }
 
 Error CommissionerImpl::Start()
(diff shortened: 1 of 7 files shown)


src/library/commissioner_impl.cpp
```diff
@@ -270,6 +271,11 @@ void CommissionerImpl::SetJoinerInfoRequester(JoinerInfoRequester aJoinerInfoReq
     mJoinerInfoRequester = aJoinerInfoRequester;
 }
 
+void CommissionerImpl::SetCommissioningHandler(CommissioningHandler aCommissioningHandler)
+{
+    mCommissioningHandler = aCommissioningHandler;
```
bukepo: I don't see where this `mCommissioningHandler` is used.

wgtdkp: Yes, the commit is coming.

wgtdkp:
Just realized that this callback call be used to support `vendor-specific provisioning` by adding those vendor-specific TLVs and ProvisioningURL TLV in JOIN_FIN.req. The return value of this callback indicates whether this joiner is accepted by application layer. like this:

```c++
using CommissioningHandler = std::function<bool(
    const JoinerInfo &aJoinerInfo,
    const std::string &aProvisioningURL,
    const std::string &aVendorName,
    const std::string &aVendorModel,
    ...,
    Error aError)>;
```

thoughts? @bukepo @librasungirl @gjc13

librasungirl:
SGTM to expose vendor-specific and provisionurl in the callback @wgtdkp

bukepo:
SGTM for include provision info.

src/library/commissioning_session.cpp
```diff
@@ -188,6 +192,22 @@ void CommissioningSession::HandleJoinFin(const coap::Request &aJoinFin)
...
+    if (mCommImpl.mCommissioningHandler != nullptr)
+    {
+        accepted = mCommImpl.mCommissioningHandler(
+            mJoinerInfo, vendorNameTlv->GetValueAsString(), vendorModelTlv->GetValueAsString(),
+            vendorSwVersionTlv->GetValueAsString(), vendorStackVersionTlv->GetValue(), provisioningUrl, vendorData);
+    }
+    else
+    {
+        // Accepts a joiner if there is no vendor-specific provisioning.
+        accepted = true;
```
bukepo: this should be `false` if provision url is provided.

PR #14, opened 2020-02-14
[tools] add commissioner THCI fixes from IOP12 and latest tests

wgtdkp:
This PR includes commissioner THCI fixes from IOP12 and latest tests:
1. set tty column to 4096. A commissioner command may exceed the tty column limit (256) set by `OpenThread_Wpanctl`;
2. Fix forever-loop in `_expect(...)` and `_command(...)`;

### DIFF
--- a/tools/commissioner_thci/commissioner_impl.py
+++ b/tools/commissioner_thci/commissioner_impl.py
@@ -143,6 +143,8 @@ class OTCommissioner(ICommissioner):
         self._handler = handler
         self._lines = []
 
+        self._command('stty cols {}'.format(TTY_COLS))
+
         config_path = '/tmp/commissioner.{}.json'.format(uuid.uuid4())
         self._write_config(config_path=config_path, config=config)
 
@@ -416,6 +418,7 @@ class OTCommissioner(ICommissioner):
                 return
 
             self._sleep(0.1)
+            timeout -= 0.1
 
         raise commissioner.Error(
             'Failed to find expected string [{}]'.format(expected))
@@ -491,6 +494,8 @@ class OTCommissioner(ICommissioner):
                 response.append(line)
 
             self._sleep(0.1)
+            timeout -= 0.1
+
         if timeout <= 0:
             raise Exception(
                 'Failed to find end of response'.format(self._port))
--- a/tools/commissioner_thci/commissioner_impl.py
+++ b/tools/commissioner_thci/commissioner_impl.py
@@ -416,9 +416,9 @@ class OTCommissioner(ICommissioner):
             if line == expected:
                 logging.info('Expected [{}]'.format(expected))
                 return
-
-            self._sleep(0.1)
-            timeout -= 0.1
+            elif line is None:
+                self._sleep(0.1)
+                timeout -= 0.1
 
         raise commissioner.Error(
             'Failed to find expected string [{}]'.format(expected))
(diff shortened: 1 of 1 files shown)


tools/commissioner_thci/commissioner_impl.py
```diff
@@ -416,6 +418,7 @@ def _expect(self, expected, timeout=10):
                 return
 
             self._sleep(0.1)
+            timeout -= 0.1
```
simonlingoogle: Would the `sleep` add unnecessary delay to the THCI? I think a better model is to only sleep when there is not a line to read yet. 
I would suggest several changes to the current code:
- `_read_line()` should only return `None` when there isn't a complete line (ends with \n) to return. I think the current implementation returns None in unnecessary conditions. 
- The code that calls `_read_line()` should not sleep if `_read_line()` successfully returns a line, but only sleep if `_read_line()` returns `None`
- Can use a smaller sleep interval: e.x. 0.01 for lower delay.

tools/commissioner_thci/commissioner_impl.py
```diff
@@ -491,6 +494,8 @@ def _command(self, cmd, timeout=10):
                 response.append(line)
 
             self._sleep(0.1)
+            timeout -= 0.1
+
```
simonlingoogle: The code look a lot like `_expect`. Why not extend `_expect` to support regression to reduce code duplication.

tools/commissioner_thci/commissioner_impl.py
wgtdkp: @Moandor-y Thoughts about returning `None`?

Moandor-y: `_read_line` will return one unnecessary `None` for each new line. To solve that we can change this function to:
```python
    def _read_line(self):
        logging.debug('Reading line')
        if len(self._lines) > 1:
            return CONTROL_SEQUENCE.sub('', self._lines.pop(0))

        tail = ''
        if len(self._lines):
            tail = self._lines.pop()

        try:
            tail += self._read()
        except socket.error:
            logging.debug('No new data')

        self._lines += NEW_LINE.split(tail)

        if len(self._lines) > 1:
            return CONTROL_SEQUENCE.sub('', self._lines.pop(0))
```

tools/commissioner_thci/commissioner_impl.py
wgtdkp: Maybe leave this to future enhancement? since this is not mandatory/related to this bugfix PR.

tools/commissioner_thci/commissioner_impl.py
```diff
@@ -431,9 +434,7 @@ def _sleep(self, seconds):
     def _read_line(self):
         logging.debug('Reading line')
         if len(self._lines) > 1:
-            line = self._lines.pop(0)
-            line = CONTROL_SEQUENCE.sub('', line)
-            return line
+            return CONTROL_SEQUENCE.sub('', self._lines.pop(0))
 
         tail = ''
         if len(self._lines):
```
simonlingoogle: nit: Can we eliminate this check by initializing `_lines` to `['']` ?

tools/commissioner_thci/commissioner_impl.py
```diff
@@ -431,9 +434,7 @@ def _sleep(self, seconds):
     def _read_line(self):
         logging.debug('Reading line')
         if len(self._lines) > 1:
-            line = self._lines.pop(0)
-            line = CONTROL_SEQUENCE.sub('', line)
-            return line
+            return CONTROL_SEQUENCE.sub('', self._lines.pop(0))
```
simonlingoogle: nit: According to how `_lines` are used in code, [`deque`](https://docs.python.org/2/library/collections.html#deque-objects) is the better choice for data structure?

wgtdkp: Agree, but consider `list` is simpler and performance is not critical to our usage, I will keep this.

PR #16, opened 2020-02-25
[code] use enum class and some constexpr

gocarlos:
* using enum class to prevent to prevent conflicts with C stuff
* adding constexpr because why not

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ build
 #tools
 .vscode
 .idea
+CMakeLists.txt.user
 
 #log
 *.log
--- a/include/commissioner/border_agent.hpp
+++ b/include/commissioner/border_agent.hpp
@@ -136,21 +136,21 @@ struct BorderAgent
 
     uint16_t mPresentFlags = 0;
 
-    static const uint16_t kAddrBit            = 1 << 0;
-    static const uint16_t kPortBit            = 1 << 1;
-    static const uint16_t kThreadVersionBit   = 1 << 2;
-    static const uint16_t kStateBit           = 1 << 3;
-    static const uint16_t kNetworkNameBit     = 1 << 4;
-    static const uint16_t kExtendedPanIdBit   = 1 << 5;
-    static const uint16_t kVendorNameBit      = 1 << 6;
-    static const uint16_t kModelNameBit       = 1 << 7;
-    static const uint16_t kActiveTimestampBit = 1 << 8;
-    static const uint16_t kPartitionIdBit     = 1 << 9;
-    static const uint16_t kVendorDataBit      = 1 << 10;
-    static const uint16_t kVendorOuiBit       = 1 << 11;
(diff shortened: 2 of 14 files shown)


.gitignore
```diff
@@ -51,6 +51,7 @@ build
 #tools
 .vscode
 .idea
+CMakeLists.txt.user
```
wgtdkp: Is this necessary?

src/library/commissioner_impl.cpp
```diff
@@ -213,7 +213,8 @@ Error CommissionerImpl::ValidateConfig(const Config &aConfig)
         VerifyOrExit(commissionerIdTlv.IsValid(), error = "invalid commissioner ID: " + aConfig.mId);
     }
 
-    VerifyOrExit(aConfig.mLogLevel <= 5, error = "invalid logging level: " + std::to_string(aConfig.mLogLevel));
+    VerifyOrExit(aConfig.mLogLevel <= LogLevel::kDebug,
+                 error = "invalid logging level: " + std::to_string(static_cast<int>(aConfig.mLogLevel)));
```
wgtdkp:
```suggestion
                 error = "invalid logging level: " + std::to_string(utils::to_underlying(aConfig.mLogLevel)));
```

include/commissioner/border_agent.hpp
```diff
@@ -136,21 +136,21 @@ struct BorderAgent
...
-    static const uint16_t kVendorNameBit      = 1 << 6;
-    static const uint16_t kModelNameBit       = 1 << 7;
-    static const uint16_t kActiveTimestampBit = 1 << 8;
-    static const uint16_t kPartitionIdBit     = 1 << 9;
-    static const uint16_t kVendorDataBit      = 1 << 10;
-    static const uint16_t kVendorOuiBit       = 1 << 11;
-    static const uint16_t kDomainNameBit      = 1 << 12;
-    static const uint16_t kBbrSeqNumberBit    = 1 << 13;
-    static const uint16_t kBbrPortBit         = 1 << 14;
+    static constexpr uint16_t kAddrBit            = 1 << 0;
```
wgtdkp: I like this. How about we do this also to constants in [network_data.hpp](https://github.com/openthread/ot-commissioner/blob/master/include/commissioner/network_data.hpp)?

.gitignore
gocarlos: yes, I work with qtcreator, this this the pendent to .vscode

gocarlos:
poh, I see that each travis ci job is compiling cmake?

gocarlos:
is it an option to do a `pip3 install cmake` to get the newer version of cmake?

wgtdkp:
> is it an option to do a `pip3 install cmake` to get the newer version of cmake?

Yes, I think it should work.

wgtdkp:
@gocarlos The travis tests failed on style checking. You need to run
```shell
./script/make-pretty.sh
```
before submitting.

PR #19, opened 2020-03-01
[script] install latest CMake with pip

wgtdkp:
This PR installs latest CMake with python3-pip rather than building from source.

### DIFF
--- a/script/bootstrap.sh
+++ b/script/bootstrap.sh
@@ -78,8 +78,8 @@ if [ $(uname) = "Linux" ]; then
 
     ## Install newest CMake
     match_version $(cmake --version | egrep -o '[0-9].*') ${MIN_CMAKE_VERSION} || {
-        pip3 install --upgrade pip
-        pip3 install cmake
+        pip3 install -U pip
+        pip3 install -U cmake
         cmake --version
     }
 
--- a/script/bootstrap.sh
+++ b/script/bootstrap.sh
@@ -45,33 +45,6 @@ match_version() {
     fi
 }
 
-## Args: $1 CMake Version. for example, "3.16.0". Reference to
-##       https://cmake.org/files/ for all available versions.
-install_cmake() {
-    local version=$1
-    local short_version=${version%.*}
-
-    if [ ! -d cmake-${version} ]; then
-        wget "https://cmake.org/files/v${short_version}/cmake-${version}.tar.gz"
-        tar -xvzf cmake-${version}.tar.gz
-    fi
-
-    local cpu_num=$(grep -c ^processor /proc/cpuinfo)
-    if [ ${cpu_num} -gt 1 ]; then
-        concurrency=$((cpu_num/2))
-    else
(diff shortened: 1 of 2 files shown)


script/bootstrap.sh
```diff
@@ -99,11 +72,14 @@ if [ $(uname) = "Linux" ]; then
...
                          ninja-build \
+                         python3-setuptools \
+                         python3-pip \
                          lcov -y
 
-    ## Install cmake 3.16.1
+    ## Install newest CMake
     match_version $(cmake --version | egrep -o '[0-9].*') ${MIN_CMAKE_VERSION} || {
-        install_cmake "3.16.1"
+        pip3 install -U pip
```
simonlingoogle: Is it necessary to update pip ?

wgtdkp: yes. otherwise, installing cmake will fail.

script/bootstrap.sh
```diff
@@ -99,11 +72,14 @@ if [ $(uname) = "Linux" ]; then
...
+        pip3 install -U cmake
```
simonlingoogle: If path is not setup correctly, users can still end up using wrong `cmake`. Should we check the `cmake` version after installation and warn the user loudly if version is not satisfied?

wgtdkp:
Depending on https://github.com/openthread/ot-commissioner/pull/20 to get travis builds passed.

script/bootstrap.sh
```diff
@@ -99,12 +72,19 @@ if [ $(uname) = "Linux" ]; then
...
+    ## Install newest CMake
+    match_version $(cmake --version | egrep -o '[0-9].*') ${MIN_CMAKE_VERSION} || {
+        pip3 install -U pip
+        pip3 install -U cmake
+    }
     match_version $(cmake --version | egrep -o '[0-9].*') ${MIN_CMAKE_VERSION} || {
-        install_cmake "3.16.1"
-        cmake --version
+        echo "error: cmake version($(cmake --version)) < ${MIN_CMAKE_VERSION}."
+        echo "did you forget to add '\$HOME/.local/bin' to beginning of your PATH?"
```
simonlingoogle:
```suggestion
        echo "Did you forget to add '\$HOME/.local/bin' to beginning of your PATH?"
```

script/bootstrap.sh
```diff
@@ -99,12 +72,19 @@ if [ $(uname) = "Linux" ]; then
...
                          cmake \
                          ninja-build \
+                         python-setuptools \
+                         python-pip \
                          lcov -y
 
-    ## Install cmake 3.16.1
+    ## Install newest CMake
+    match_version $(cmake --version | egrep -o '[0-9].*') ${MIN_CMAKE_VERSION} || {
+        pip install -U pip
```
wgtdkp: The integration test uses OT Commissioner THCI which requires python2 by default. We here stick to python2 and see if we can upgrade to python3 in a future PR.

PR #22, opened 2020-03-02
[build] allow downstream to change the default build flags

gocarlos:


CMakeLists.txt
```diff
@@ -29,18 +29,24 @@
...
+    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    set(CMAKE_CXX_EXTENSIONS OFF)
+endif()
 
-set(CMAKE_CXX_STANDARD 11)
-set(CMAKE_CXX_STANDARD_REQUIRED ON)
-set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_CXX_FLAGS "-Wall -Wextra -Werror -Wfatal-errors -Wno-missing-braces")
+if(OT_COMM_MAINTAINER_FLAGS)
```
wgtdkp: In which case will this option be useful? I think those FLAGS has effect on only OT Commissioner project and dependencies.

gocarlos: E.g. using yocto: the conpiler flags are set globally for all libraries and executables... this was resetting those settings

wgtdkp: you meaning using yocto as a library of OT Commissioner?

gocarlos: yocto is a build system to create a linux distribution. this system is widely used to create costum industrial grade linux-based OSs.

Yocto relies on recipes which use the native build system e.g. cmake to install artifacts to the rootfs of the OS image.

yocto will use different flags depending on the machine/architecture/rootfs etc.

those flags are passed to the recipes and build systems e.g. cmake, autotools, scons, meson etc.

Those flags have to be respected by the libraries in order to have a common ABI and link different libs and executables together...

wgtdkp: >Those flags have to be respected by the libraries in order to have a common ABI and link different libs and executables together...

it seems current `CMAKE_CXX_FLAGS` specifies only warning strategies. For your requirement of passing extra FLAGS to OT Commissioner, how about we set `CMAKE_CXX_FLAGS` as:
```cmake

```set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra -Werror -Wfatal-errors -Wno-missing-braces")

Comment (gocarlos) [reply CMakeLists.txt]: top

Comment (gocarlos) [reply CMakeLists.txt]: so, but this means no support for msvc?

Comment (wgtdkp) [reply CMakeLists.txt]: We current don't support Windows. May improve later to support Windows.

File: CMakeLists.txt [hunk 2]
```diff
@@ -29,18 +29,21 @@
...
     set(CMAKE_BUILD_TYPE Debug)
 endif()
 
-option(OT_COMM_COVERAGE "Enable coverage reporting" OFF)
-option(OT_COMM_TEST     "Build tests" ON)
-option(OT_COMM_APP      "Build the CLI App" ON)
+if(NOT "${CMAKE_CXX_STANDARD}")
+    set(CMAKE_CXX_STANDARD 11)
+    set(CMAKE_CXX_STANDARD_REQUIRED ON)
+    set(CMAKE_CXX_EXTENSIONS OFF)
```
wgtdkp: This doesn't seem to be related to CXX_STANDARD. Can we move it out?

PR #25, opened 2020-03-03
[tests] enhance integration test setup for OTBR

wgtdkp:
This PR enhances integration test for OTBR by
- provide an environment to skip building OTBR from source.
- reduce git clone by option `--depth=1`.
- change tmp dir to `/tmp/test-ot-commissioner` to keep current dir clean.
- Add file descriptions.

Those changes enable reusing the integration test script in ot-br-posix project.

### DIFF
--- a/tests/integration/bootstrap.sh
+++ b/tests/integration/bootstrap.sh
@@ -105,7 +105,7 @@ main() {
     set -e
     mkdir -p ${RUNTIME_DIR}
 
-    if [[ $SKIP_BUILDING_OTBR == 0 ]] && [ ! -d ${OTBR} ]; then
+    if (( $SKIP_BUILDING_OTBR == 0 )) && [ ! -d ${OTBR} ]; then
         setup_otbr
     fi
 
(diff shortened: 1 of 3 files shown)


tests/integration/bootstrap.sh
```diff
@@ -27,11 +27,20 @@
 #  POSSIBILITY OF SUCH DAMAGE.
 #
 
+## This file bootstrap dependencies of integration tests.
```
bukepo: why double `##`?

tests/integration/bootstrap.sh
```diff
@@ -96,7 +105,7 @@ main() {
     set -e
     mkdir -p ${RUNTIME_DIR}
 
-    if [ ! -d ${OTBR} ]; then
+    if [[ $SKIP_BUILDING_OTBR == 0 ]] && [ ! -d ${OTBR} ]; then
```
bukepo: Is mixing `[[ ]]` and `[]` intended?

wgtdkp: I actually want arithmetic comparison here, changed to `(())`.

tests/integration/bootstrap.sh
wgtdkp: Not really necessary, just to mimic c-like languages' single line comment(`//`).

Issue #27, opened 2020-03-11
`kTransportFailed` error is generated in response to a bad PSKc


jwhui:
An error like `kSecurity` may be more appropriate in this case.

wgtdkp:
Agree.

PR #33, opened 2020-03-12
[dtls] refine DTLS state and error handling

wgtdkp:
This PR refines DTLS state and error handling:

- return `Error::kSecurity` for DTLS errors (return `Error::kTransportFailed` for net error)
- unify error casting from mbedtls result
- remove `State::kDisconnecting` state to simplify state transition

This PR addresses [issue#27](https://github.com/openthread/ot-commissioner/issues/27).

### DIFF
--- a/src/library/dtls.cpp
+++ b/src/library/dtls.cpp
@@ -87,29 +87,39 @@ static void HandleMbedtlsDebug(void *, int level, const char *file, int line, co
  * For the implementation details, please reference to <mbedtls/error.h>.
  *
  */
-static Error ErrorFromMbedtlsError(int mbedtlsError)
+static Error ErrorFromMbedtlsError(int aMbedtlsError)
 {
-    ASSERT(mbedtlsError <= 0);
+    // See <mbedtls/error.h> for the constants.
+    static constexpr int kMbedtlsErrorLowLevelNetBegin        = -0x0052;
+    static constexpr int kMbedtlsErrorLowLevelNetEnd          = -0x0042;
+    static constexpr int kMbedtlsErrorHighLevelModuleIdMask   = 0x7000;
+    static constexpr int kMbedtlsErrorHighLevelModuleIdOffset = 12;
+    static constexpr int kMbedtlsErrorHighLevelModuleIdCipher = 6;
+    static constexpr int kMbedtlsErrorHighLevelModuleIdSsl    = 7;
+
+    ASSERT(aMbedtlsError <= 0);
 
     Error error;
 
-    uint16_t highLevelModuleId = (static_cast<uint16_t>(-mbedtlsError) >> 12) & 0x07;
+    uint16_t highLevelModuleId = (static_cast<uint16_t>(-aMbedtlsError) & kMbedtlsErrorHighLevelModuleIdMask) >>
+                                 kMbedtlsErrorHighLevelModuleIdOffset;
 
-    if (mbedtlsError == 0)
+    if (aMbedtlsError == 0)
     {
         error = Error::kNone;
     }
-    else if (mbedtlsError == MBEDTLS_ERR_SSL_WANT_READ || mbedtlsError == MBEDTLS_ERR_SSL_WANT_WRITE ||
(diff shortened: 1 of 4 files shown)


src/library/dtls.cpp
```diff
@@ -80,6 +81,47 @@ static void HandleMbedtlsDebug(void *, int level, const char *file, int line, co
     }
 }
 
+/**
+ * This function convert mbedtls error to OT Commissioner error.
+ *
+ * For the implementation details, please reference to <mbedtls/error.h>.
+ *
+ */
+static Error ErrorFromMbedtlsError(int mbedtlsError)
```
bukepo:
```suggestion
static Error ErrorFromMbedtlsError(int aMbedtlsError)
```

src/library/dtls.cpp
```diff
@@ -80,6 +81,47 @@ static void HandleMbedtlsDebug(void *, int level, const char *file, int line, co
...
+    if (mbedtlsError == 0)
+    {
+        error = Error::kNone;
+    }
+    else if (mbedtlsError == MBEDTLS_ERR_SSL_WANT_READ || mbedtlsError == MBEDTLS_ERR_SSL_WANT_WRITE ||
+             mbedtlsError == MBEDTLS_ERR_SSL_ASYNC_IN_PROGRESS || mbedtlsError == MBEDTLS_ERR_SSL_CRYPTO_IN_PROGRESS)
+    {
+        error = Error::kTransportBusy;
+    }
+    else if (mbedtlsError <= -0x0042 && mbedtlsError >= -0x0052)
```
bukepo: is it possible to avoid these magic numbers and those in line 112?

wgtdkp: I don't think so. There are no macros indicating the begin and end of the error code of specific category, for example `MBEDTLS_SSL_ERROR_BEGIN` and `MBEDTLS_SSL_ERROR_END`. This could be a PR to mbedtls, but at least I can define a local macro for our usage. Will update.

src/library/dtls.cpp
```diff
@@ -80,6 +81,57 @@ static void HandleMbedtlsDebug(void *, int level, const char *file, int line, co
     }
 }
 
+/**
+ * This function convert mbedtls error to OT Commissioner error.
+ *
+ * For the implementation details, please reference to <mbedtls/error.h>.
+ *
+ */
+static Error ErrorFromMbedtlsError(int aMbedtlsError)
```
simonlingoogle: Is it possible to preserve the original `aMbedtlsError` in Error?

src/library/dtls.cpp
```diff
@@ -360,47 +404,48 @@ void DtlsSession::HandleEvent(short aFlags)
...
-    VerifyOrExit(rval == 0, error = Error::kOutOfMemory);
+    SuccessOrExit(error = ErrorFromMbedtlsError(rval));
 
     mIsClientIdSet = true;
 
 exit:
     return error;
 }
 
+bool DtlsSession::ShouldStop(Error aError)
```
simonlingoogle: Do you think its better to define a member method like `bool Error::IsTemporary()` or `bool Error::IsFatal()` and call it here?

src/library/dtls.cpp
```diff
@@ -360,47 +404,48 @@ void DtlsSession::HandleEvent(short aFlags)
...
+{
+    return aError != Error::kNone && aError != Error::kTransportBusy;
```
simonlingoogle:
```suggestion
    return aError != Error::kNone && !aError.IsTemporary();
```
as I said above.

src/library/dtls.cpp
wgtdkp: I don't think it is easy to implement with refactoring the `Error` class. By the way, we have already refactored `Error` to include a error message and in the case of mbedtls, it includes the string form of `aMbedtlsError`.

src/library/dtls.cpp
wgtdkp: The error code is generic, whether an error code is fatal or not may differ in different context. So I think it is not applicable at here. thoughts?

PR #34, opened 2020-03-12
[error] make error unignorable

wgtdkp:
This PR makes error unignorable by applying `warn_unused_result` attribute. 

When a type is annotated with `warn_unused_result` attribute, any function returns the object of this type must explicitly check(or ignore) the return value to get rid of the warning/error.

OT mimics exception with two error handling macros: `verifyOrExit` and `SuccessorExit`. But it is easy for a function calling to ignore the return error silently, resulting in escaped errors. And, when looking at a calling without checking the return error, the reader doesn't know if ignoring error is intentional or because of careless.

I think it is crutial to always explicitly handle (check or ignore) a return error, as Google's famous `Status` class (something like `Error` class here) is not ignorable.

We achieve this by:
- annotate `Error` class with the `warn_unused_result` attribute.
- provide macro `IgnoreError` to explicitly ignore an error when ignoring it is the desired behavior.

This approach can also be migrated to openthread and ot-br-posix.

### DIFF
--- a/include/commissioner/error.hpp
+++ b/include/commissioner/error.hpp
@@ -36,11 +36,13 @@
 
 #include <string>
 
+#include <commissioner/defines.hpp>
+
 namespace ot {
 
 namespace commissioner {
 
-enum class Error : int
+enum class OT_COMM_MUST_USE_RESULT Error : int
 {
     kNone = 0,        ///< No error.
     kAbort,           ///< A request is aborted.
--- a/src/app/commissioner_app.cpp
+++ b/src/app/commissioner_app.cpp
@@ -155,7 +155,7 @@ exit:
 
 void CommissionerApp::Stop()
 {
-    mCommissioner->Resign();
+    IgnoreError(mCommissioner->Resign());
 }
 
 void CommissionerApp::AbortRequests()
@@ -1340,18 +1340,19 @@ void CommissionerApp::HandleEnergyReport(const std::string *aPeerAddr,
                                          const ByteArray *  aEnergyList,
                                          Error              aError)
 {
-    if (aError != Error::kNone)
-    {
-        return;
-    }
-
+    Error   error;
     Address addr;
-    addr.Set(*aPeerAddr);
-    ASSERT(addr.IsValid());
+
+    SuccessOrExit(error = aError);
+
+    SuccessOrExit(error = addr.Set(*aPeerAddr));
 
     // Main thread will wait for updates to mPanIdConflicts,
     // which guarantees no concurrent access to it.
     mEnergyReports[addr] = {*aChannelMask, *aEnergyList};
+
+exit:
+    return;
 }
 
 void CommissionerApp::HandleDatasetChanged(Error error)
--- a/src/app/commissioner_app.cpp
+++ b/src/app/commissioner_app.cpp
@@ -1340,12 +1340,10 @@ void CommissionerApp::HandleEnergyReport(const std::string *aPeerAddr,
                                          const ByteArray *  aEnergyList,
                                          Error              aError)
 {
-    Error   error;
     Address addr;
 
-    SuccessOrExit(error = aError);
-
-    SuccessOrExit(error = addr.Set(*aPeerAddr));
+    SuccessOrExit(aError);
+    SuccessOrExit(addr.Set(*aPeerAddr));
 
     // Main thread will wait for updates to mPanIdConflicts,
     // which guarantees no concurrent access to it.
--- a/src/common/address.cpp
+++ b/src/common/address.cpp
@@ -105,7 +105,8 @@ exit:
 Address Address::FromString(const std::string &aAddr)
 {
     Address ret;
-    ret.Set(aAddr);
+    auto    error = ret.Set(aAddr);
+    ASSERT(error == Error::kNone);
     return ret;
 }
 
--- a/src/library/coap.cpp
+++ b/src/library/coap.cpp
@@ -82,8 +82,6 @@ Error Message::AppendOption(OptionType aNumber, const OptionValue &aValue)
 {
     Error error = Error::kNone;
 
-    // This will silently ignore options with unrecognized
-    // option number and elective options with bad value.
     VerifyOrExit(IsValidOption(aNumber, aValue), error = Error::kInvalidArgs);
 
     if (aNumber == OptionType::kUriPath)
@@ -514,7 +512,7 @@ void Coap::HandleRequest(const Request &aRequest)
     }
     else
     {
-        SendNotFound(aRequest);
+        IgnoreError(SendNotFound(aRequest));
         ExitNow(error = Error::kNotFound);
     }
 
@@ -535,7 +533,7 @@ void Coap::HandleResponse(const Response &aResponse)
     {
         if (aResponse.IsConfirmable() || aResponse.IsNonConfirmable())
         {
-            SendReset(aResponse);
+            IgnoreError(SendReset(aResponse));
         }
         ExitNow();
     }
@@ -578,7 +576,7 @@ void Coap::HandleResponse(const Response &aResponse)
 
     case Type::kConfirmable:
         // Send empty ACK if it is a CON message.
-        SendAck(aResponse);
+        IgnoreError(SendAck(aResponse));
 
         // fall through
 
@@ -996,11 +994,30 @@ std::shared_ptr<Message> Message::Deserialize(Error &aError, const ByteArray &aB
 
         SuccessOrExit(error = Deserialize(number, value, lastOptionNumber, aBuf, offset));
 
-        // Stop if any unrecognized option is critical.
-        // Otherwise, we need to proceed to decoding following options.
-        VerifyOrExit(IsValidOption(number, value) || !IsCriticalOption(number), error = Error::kBadOption);
+        if (IsValidOption(number, value))
+        {
+            // AppendOption will do further validation before adding the option.
+            error = message->AppendOption(number, value);
+        }
+        else
+        {
+            error = Error::kBadFormat;
+        }
 
-        message->AppendOption(number, value);
+        if (error != Error::kNone)
+        {
+            if (IsCriticalOption(number))
+            {
+                // Stop if any unrecognized option is critical.
+                ExitNow(error = Error::kBadOption);
+            }
+            else
+            {
+                // Ignore non-critical option error.
+                error = Error::kNone;
+            }
+        }
+
         lastOptionNumber = utils::to_underlying(number);
     }
 
--- a/src/library/coap.cpp
+++ b/src/library/coap.cpp
@@ -610,7 +610,7 @@ void Coap::Retransmit(Timer &)
             mRequestsCache.Put(requestHolder);
 
             std::string uri;
-            requestHolder.mRequest->GetUriPath(uri);
+            IgnoreError(requestHolder.mRequest->GetUriPath(uri));
 
             // Retransmit
             if (!requestHolder.mAcknowledged)
--- a/src/library/commissioner_impl.cpp
+++ b/src/library/commissioner_impl.cpp
@@ -161,11 +161,11 @@ CommissionerImpl::CommissionerImpl(struct event_base *aEventBase)
     , mJoinerInfoRequester(nullptr)
     , mCommissioningHandler(nullptr)
 {
-    mBrClient.AddResource(mResourceUdpRx);
-    mBrClient.AddResource(mResourceRlyRx);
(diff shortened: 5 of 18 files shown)


src/app/commissioner_app.cpp
```diff
@@ -152,7 +152,7 @@ Error CommissionerApp::Start(const std::string &aBorderAgentAddr, uint16_t aBord
 
 void CommissionerApp::Stop()
```
simonlingoogle: let Stop() return error ?

src/app/commissioner_app.cpp
```diff
@@ -1337,18 +1337,19 @@ void CommissionerApp::HandleEnergyReport(const std::string *aPeerAddr,
...
-    {
-        return;
-    }
-
+    Error   error;
     Address addr;
-    addr.Set(*aPeerAddr);
-    ASSERT(addr.IsValid());
+
+    SuccessOrExit(error = aError);
```
simonlingoogle:
```suggestion
    SuccessOrExit(aError);
```
Is it ok?

src/common/address.cpp
```diff
@@ -105,7 +105,8 @@ Error Address::ToString(std::string &aAddr) const
 Address Address::FromString(const std::string &aAddr)
 {
     Address ret;
-    ret.Set(aAddr);
+    auto    error = ret.Set(aAddr);
+    ASSERT(error == Error::kNone);
```
simonlingoogle: Is `assert` too strict for parsing error?

src/library/commissioner_impl.cpp
```diff
@@ -161,11 +161,11 @@ CommissionerImpl::CommissionerImpl(struct event_base *aEventBase)
     , mJoinerInfoRequester(nullptr)
     , mCommissioningHandler(nullptr)
 {
-    mBrClient.AddResource(mResourceUdpRx);
-    mBrClient.AddResource(mResourceRlyRx);
-    mProxyClient.AddResource(mResourceDatasetChanged);
-    mProxyClient.AddResource(mResourcePanIdConflict);
-    mProxyClient.AddResource(mResourceEnergyReport);
+    ASSERT(mBrClient.AddResource(mResourceUdpRx) == Error::kNone);
```
simonlingoogle: Define another macro to simplify these codes?
```suggestion
    MustSuccess(mBrClient.AddResource(mResourceUdpRx));
```

src/library/socket.cpp
```diff
@@ -184,7 +184,7 @@ Address UdpSocket::GetLocalAddr() const
     ASSERT(getsockname(mNetCtx.fd, reinterpret_cast<sockaddr *>(&addr), &len) == 0);
 
     Address ret;
-    ret.Set(addr);
+    ASSERT(ret.Set(addr) == Error::kNone);
```
simonlingoogle: Is this ASSERT and the one above too strict?

src/app/commissioner_app.cpp
wgtdkp: I think we can always force stop a Commissioner.

src/common/address.cpp
wgtdkp: - The semantic of `FromString` is `don't accept invalid aAddr`.
- Each usage of `FromString` is passed a string literal, So we are rejecting programming bug.
- FromString is internally accessible.

src/library/commissioner_impl.cpp
wgtdkp: good idea. I have the same idea in my head, but leave it to future PR.

src/library/socket.cpp
wgtdkp: Yes, I agree. We have an assumption that the this function will only be called when the socket is in valid state (actually, this function is designed for unit testing). But we should no have this assumption.

wgtdkp: Added comment for the methods (https://github.com/openthread/ot-commissioner/pull/34/commits/f2d5a049ad15c0141e5a1e34d608ed4061fe69a9). please take a look if it is ok with you. A perfect handling requires changes to the API and its semantic, may leave it in future PR.

include/commissioner/error.hpp
```diff
@@ -36,11 +36,13 @@
...
 #include <string>
 
+#include <commissioner/defines.hpp>
+
 namespace ot {
 
 namespace commissioner {
 
-enum class Error : int
+enum class OT_COMM_MUST_USE_RESULT Error : int
```
abtink: Do you happen to know if we use the same `__attribute__((warn_unused_result))` with `enum` and not `enum class`?

src/app/commissioner_app.cpp
```diff
@@ -152,7 +152,7 @@ Error CommissionerApp::Start(const std::string &aBorderAgentAddr, uint16_t aBord
 
 void CommissionerApp::Stop()
 {
-    mCommissioner->Resign();
+    IgnoreError(mCommissioner->Resign());
```
abtink: We have a `IgnoreReturnValue()` macro.  Can we consider using/combining with that?

src/app/commissioner_app.cpp
abtink: I agree. I don't we need the `error` variable here on next line since we are not using it after this.
```
SuccessOrExit(addr.Set(*aPeerAddr));
```

src/app/commissioner_app.cpp
abtink: Just noticed we don't have this `IgnoreReturnValue()` in this repo. It is in OT main.

include/commissioner/error.hpp
wgtdkp: Annotating class/enum with `war_unused_result` is supported by clang. It has not difference between `enum` and `enum class`. I was trying to apply this to `openthread/openthread`, but decided to try it on this repo first to see how it works and if we like it.

src/app/commissioner_app.cpp
wgtdkp: even in repo `openthread/openthread`, I think `IgnoreError` makes it clearer that we are ignoring  a error. Of cause, we can define `IgnoreError` to `IgnoreReturnValue`.

src/library/coap.cpp
```diff
@@ -995,11 +993,30 @@ std::shared_ptr<Message> Message::Deserialize(Error &aError, const ByteArray &aB
...
+            {
+                // Stop if any unrecognized option is critical.
+                ExitNow(error = Error::kBadOption);
+            }
+            else
+            {
+                // Ignore non-critical option error.
+                error = Error::kNone;
+            }
+        }
```
Moandor-y: If `IsValidOption(number, value) == false` and `IsCriticalOption(number) == false`, `message->AppendOption(number, value)` would be executed before, but not after this change.

Is this change of behavior intended?

wgtdkp: @Moandor-y thanks for you careful consideration. Actually, `AppendOption` itself does arguments validation and rejects if `IsValidOption(number, value) == false`. Previous implementation ignores the error returned by `AppendOption` silently. This PR should not change the actual behavior but explicitly ignore the error by setting it to `Error::kNone`.

src/app/commissioner_app.cpp
jwhui: In some cases, I believe `IgnoreReturnValue` is more descriptive. For example, in cases where the error is passed via a parameter. Thoughts?

wgtdkp: @jwhui, Yes `IgnoreError` is not to be used for any return values. It is designed to be used for `Error` / `otError` exclusively.

jwhui: I think `IgnoreReturnValue` is more descriptive of its intended purpose - to ignore the return value. It's not clear to me why we need to define a separate macro.

wgtdkp: What I am thinking is:

Typically, there are two kinds of return value: error code and normal data. When readers see `IgnoreError(...)`, they know we are ignoring any error of a function call and we allow errors at there. While `IgnoreReturnValue(...)` doesn't give the information of what is being ignored (error or normal data), so the reader doesn't know if we allow a failure at there untill he checks the function signature by himself.

Current usages of `IgnoreReturnValue` in openthread are all applied on functions return `otError`, **I think, because only error is something deserves an explicit ignoring**. For example, `Message::GetLength()` always success and return the length of the message, there is no meaning for the caller to explicitly ignore the return value, because if the return value is not needed, why should the function be called? 

There are third-party libraries that return error code and normal data in a single signed integer, but I don't think it is a good practice in openthread code. If we need to check the return value of a function against some constants or boundaries to verify if the function call succeed, then we should make the function return `otError` and pass the "true" return value through a by-reference parameter. While we can still use `IgnoreReturnValue` on those thrid-party APIs.

Thoughts?

abtink: I would also favor using one (more generic) macro `IgnoreReturnValue`.

> Current usages of IgnoreReturnValue in openthread are all applied on functions return otError, I think, because only error is something deserves an explicit ignoring.

I think we already have cases where we want to ignore return values that are not necessarily an error.  Example of this is `Message::Read()` which returns number of bytes read from a message. If the developer knows already that the number of bytes that are being read is smaller than available length in the message, it is perfectly reasonable to ignore the return value.  Example of this in [mesh_forwarder_ftd.cpp](https://github.com/openthread/openthread/blob/7456be9e80b230b734ad454d13243b2978e3364e/src/core/thread/mesh_forwarder_ftd.cpp#L238-L250).

wgtdkp: @abtink Thanks for your example. My overall idea is making `Error` and `otError` special, but not generic (**failure and return value are totally two different things I think**). In high-level languages, they use `exception` for failures and `exception` is not ignorable. I here want to mimic the semantic of `exception` with `otError` and make it also not ignorable (since we have already do so with macros `VerifyOrExit` and `SuccessOrExit`). To force using `IgnoreError` on only `otError`, we can even have it as a function:
```c++
static inline void IgnoreError(otError aError)
{
    (void)aError;
}
```

Google doesn't allow exception in C++ code and we have [abseil::Status](https://github.com/abseil/abseil-cpp/blob/master/absl/status/status.h) (abseil is the opensource version of the C++ fundamental library used inside Google) to represent failures (or success).

@jwhui @abtink If you agree on making `otError` special than general return value, I think `IgnoreError` is better than `IgnoreReturnValue`. Otherwise, I am still Ok with using `IgnoreReturnValue`.

simonlingoogle: `otError` and `Error` are used extensively in `openthread` and `ot-commissioner` code, making it worthy to be treated differently from other return values. 
I think using `IgnoreError` has following advantages over using `IgnoreReturnValue`:

- `IgnoreError` tells the code reader explicitly that what we are ignoring here is an `Error`, otherwise the reader could wonder what's the type of the return value that is ignored here? He/she will have to check the prototype of the called function to make it clear.
- In the case when `otError` is passed as a function argument, we can also use `IgnoreError` instead of `OT_UNUSED_VARIABLE` to explicit convey our intention to ignore the `Error`. 
- `IgnoreError` is shorter than `IgnoreReturnValue`.

Overall, I think the benefits of using `IgnoreError` outweigh the cost.

abtink: I fully agree not allowing returned `otError` from methods/function to be ignored would be a useful change. It ensures that developer to consider each situation (either handle the error or decide it is safe to ignore it). 

Having an `IgnoreError()` is fine with me, but we still need to have `IgnoreReturnValue` for other cases (where return value is not `otError` and would be useful for developer to indicate she/he is purposely ignoring it).

One thing to be careful is that  `otError` use in OT does not fully equate to the notion of exceptions. Exceptions typically indicate  an unexpected/rare failure (in bad or unrecoverable way) and they are not expected generally to happen as part of the normal flow of a program. The `otError` enumeration on other hand is used to convey "status" in many cases. Getting an `otError` (which is not an `OT_ERROR_NONE`) in many situations is common or even part of the normal flow of the code. Good examples of this are methods/APIs that iterate over some list and they return `OT_ERROR_NOT_FOUND` when iterator reaches end of the list. Getting this error  is expected as part of the normal flow when going through the list.

wgtdkp: @abtink Thanks for your explanation of `otError` usage! Agree that we cannot take `otError` the same thing as an Exception since normal flow may depend on the error code. As `otError` is the only representation of failure (at least it should be?) and in most cases in represents for only failure (of cause it can be used for normal flow in specific cases), I believe it deserves a separate macro/function to handler it, as either an return value or function parameter.

wgtdkp:
@abtink @jwhui more comments?

PR #35, opened 2020-03-14
[code] avoid static objects with non-trivial destructor

wgtdkp:
This PR removes eliminates static objects with non-trivial destructor. This is to resolve the out-of-order destruction issue of those static object which may leading to bad access to already destructed objects.

### DIFF
--- a/src/library/uri.hpp
+++ b/src/library/uri.hpp
@@ -45,45 +45,45 @@ namespace uri {
 /*
  * Thread MeshCoP URIs
  */
-static const char *kPetitioning         = "/c/cp";
-static const char *kKeepAlive           = "/c/ca";
-static const char *kUdpRx               = "/c/ur";
-static const char *kUdpTx               = "/c/ut";
-static const char *kRelayRx             = "/c/rx";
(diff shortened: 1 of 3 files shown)


src/library/uri.hpp
```diff
@@ -45,45 +45,45 @@ namespace uri {
...
-static const std::string kMgmtEdScan          = "/c/es";
-static const std::string kMgmtEdReport        = "/c/er";
-static const std::string kMgmtReenroll        = "/c/re";
-static const std::string kMgmtDomainReset     = "/c/rt";
-static const std::string kMgmtNetMigrate      = "/c/nm";
-static const std::string kMgmtSecPendingSet   = "/c/sp";
-static const std::string kJoinEnt             = "/c/je";
-static const std::string kJoinFin             = "/c/jf";
-static const std::string kJoinApp             = "/c/ja";
+static const char *kPetitioning         = "/c/cp";
```
Moandor-y: Consider changing the type to `const char* const` to avoid modifying the pointer itself.

wgtdkp: done.

PR #38, opened 2020-03-17
[script] sync make-pretty scripts from openthread

jwhui:
- Also reformat files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,8 +27,7 @@
 #
 
 language: c++
-
-script: .travis/script.sh
+dist: bionic
 
 after_success: .travis/after_success.sh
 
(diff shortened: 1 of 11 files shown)


script/clang-format.sh
```diff
@@ -1,6 +1,6 @@
-#!/bin/bash
+#!/bin/sh
```
wgtdkp: nit: Other scripts use `bash`. Should we unify which shell we use? Inside Google, `bash` is chosen for all shell scripts :)

bukepo: this script requires bash.

jwhui: I am fine with bash. I'll go ahead and make the updates, including to openthread/openthread.

Issue #40, opened 2020-03-18
[error] include detailed message in error.


wgtdkp:
We want more user readable information about an error, just as an `exception` provides. So we gonna need to include a descriptive message when an error is returned since exception is disallowed in Google (that's sad).

We here propose to mimic google's famous `Status` class which includes a message when there is an error :

https://github.com/abseil/abseil-cpp/blob/7853a7586c492ce8905c9e49f8679dada6354f2c/absl/status/status.h#L75

PR #45, opened 2020-03-24
[error] add error message to the Error class.

wgtdkp:
This PR adds readable message to `Error` class and requires the message when creating an error. The error message provides much more descriptive information to user about the failure, and will also be useful for logging and debugging.

**Change summary:**
- Refactor `Error` to include an error message ([error.hpp](https://github.com/openthread/ot-commissioner/pull/45/files#diff-a5ab74458d54760fae04d4326b760984))
- Must specify an error message when returning an error, example:
```c++
VerifyOrExit((aDataset.mPresentFlags & CommissionerDataset::kBorderAgentLocatorBit) == 0,
             error = ERROR_INVALID_ARGS("trying to set Border Agent Locator which is read-only"));
```
- Detailed DTLS error with `mbedtls_strerror` ([mbedtls-error](https://github.com/openthread/ot-commissioner/pull/45/files#diff-509885766818e5b5ad7fecd493ffd065))

Please take a look about the [Error](https://github.com/openthread/ot-commissioner/pull/45/files#diff-a5ab74458d54760fae04d4326b760984) class and how the code is like after adding message to Error.

**Addresses issue https://github.com/openthread/ot-commissioner/issues/40.**

**Related PRs**
- [Make error unignorable](https://github.com/openthread/ot-commissioner/pull/34)
- [Refine DTLS error](https://github.com/openthread/ot-commissioner/pull/33), The changes here include this PR and will do rebase once it got merged.


wgtdkp:
This PR is now ready for review, @bukepo @librasungirl @abtink @simonlingoogle please have a look.

include/commissioner/error.hpp
```diff
@@ -28,48 +28,224 @@
...
+    Error(const Error &aError);
+    Error &operator=(const Error &aError);
+    Error(Error &&aError) noexcept;
+    Error &operator=(Error &&aError) noexcept;
+
+    /**
+     * Returns true if there is no error.
+     *
+     */
+    bool NoError() const { return (mState == nullptr); }
```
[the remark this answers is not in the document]
wgtdkp: what about `IsNone`? `error.IsSuccess()` is weird.

include/commissioner/error.hpp
```diff
@@ -42,36 +43,211 @@ namespace ot {
...
+     *
+     */
+    Error() {}
+
+    /**
+     * Creates a error with the specified error code and message as a
+     * human-readable string containing more detailed information.
+     *
+     */
+    Error(ErrorCode aErrorCode, std::string aErrorMessage);
```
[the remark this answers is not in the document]
wgtdkp: The `aErrorMessage` has to be copied at least once anyway. I am okay to switch to `pass by reference`.

include/commissioner/error.hpp
```diff
@@ -42,36 +43,211 @@ namespace ot {
...
+     * Returns true if there is no error.
+     *
+     */
+    bool NoError() const { return (mState == nullptr); }
+
+    ErrorCode GetCode() const { return NoError() ? ErrorCode::kNone : mState->mCode; }
+
+    const std::string &GetMessage() const { return NoError() ? EmptyString() : mState->mMessage; }
+
+    void SetMessage(const std::string &aMessage);
```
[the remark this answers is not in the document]
wgtdkp: Agree.

src/app/border_agent.cpp
```diff
@@ -83,13 +84,13 @@ Error DiscoverBorderAgent(BorderAgentHandler aBorderAgentHandler, size_t aTimeou
...
-        if (!curBorderAgentOrErrorMsg.mErrorMsg.empty())
+        if (!curBorderAgentOrErrorMsg.mError.NoError())
         {
-            aBorderAgentHandler(nullptr, &curBorderAgentOrErrorMsg.mErrorMsg);
+            aBorderAgentHandler(nullptr, curBorderAgentOrErrorMsg.mError);
         }
         else if (curBorderAgentOrErrorMsg.mBorderAgent.mPresentFlags != 0)
         {
-            aBorderAgentHandler(&curBorderAgentOrErrorMsg.mBorderAgent, nullptr);
+            aBorderAgentHandler(&curBorderAgentOrErrorMsg.mBorderAgent, ERROR_NONE);
```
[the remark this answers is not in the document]
wgtdkp: `No Border Agents Found` is not treated as an error.

src/app/border_agent.cpp
```diff
@@ -361,7 +368,7 @@ static int HandleRecord(const struct sockaddr *from,
     }
 
 exit:
-    return error == Error::kNone ? 0 : -1;
+    return 0;
```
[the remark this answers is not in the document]
wgtdkp: The return value is used by the `mDNS` library (please notice that we are in a callback of `mDNS`). `mDNS` stops parsing DNS records if this function returns non-zero value. The `error` here represents logic errors specific to Thread protocol, so we should derived the return value from `error`, our previous implementation is considered not correct.

We tolerate malformed `mDNS` records and return "no border agent".

src/library/token_manager.cpp
```diff
@@ -442,12 +453,14 @@ Error TokenManager::ParsePublicKey(mbedtls_pk_context &aPublicKey, const ByteArr
...
-    VerifyOrExit(!aPrivateKeyRaw.empty());
-    VerifyOrExit(mbedtls_pk_parse_key(&aPrivateKey, &aPrivateKeyRaw[0], aPrivateKeyRaw.size(), nullptr, 0) == 0);
+    Error error;
 
-    error = Error::kNone;
+    VerifyOrExit(!aPrivateKeyRaw.empty(), error = ERROR_INVALID_ARGS("the raw private key is empty"));
+    if (int fail = mbedtls_pk_parse_key(&aPrivateKey, aPrivateKeyRaw.data(), aPrivateKeyRaw.size(), nullptr, 0))
+    {
+        error = ErrorFromMbedtlsError(fail);
+        ExitNow(error = {ErrorCode::kInvalidArgs, error.GetMessage()});
```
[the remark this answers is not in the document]
wgtdkp: The root cause is described by the `error message`. The `error code` just gives the category of this error and whatever category the underlying error belongs to, in the context of `ParsePrivateKey`, it conveys `Invalid Arguments`.

src/library/socket.cpp
```diff
@@ -210,7 +210,7 @@ Address UdpSocket::GetPeerAddr() const
     ASSERT(getpeername(mNetCtx.fd, reinterpret_cast<sockaddr *>(&addr), &len) == 0);
 
     Address ret;
-    ASSERT(ret.Set(addr) == Error::kNone);
+    ASSERT(ret.Set(addr).NoError());
```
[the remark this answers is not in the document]
wgtdkp: `ASSERT` in OT Commissioner has its implementation that it is not removed in release build so it can have side effect, this is why it is defined.

I am okay to change the behavior of `ASSERT` to follow `assert` and avoid side effect.

src/common/address.cpp
```diff
@@ -76,37 +97,39 @@ Error Address::Set(const sockaddr_storage &aSockAddr)
...
-    return error;
+    return result;
 }
 
 Address Address::FromString(const std::string &aAddr)
 {
     Address ret;
     auto    error = ret.Set(aAddr);
-    ASSERT(error == Error::kNone);
+    ASSERT(error.NoError());
```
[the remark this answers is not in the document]
wgtdkp: This is for only test.

include/commissioner/error.hpp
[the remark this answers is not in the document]
simonlingoogle: I would still prefer `IsSuccess()`, but `IsNone` is fine to me.

include/commissioner/error.hpp
```diff
@@ -42,36 +43,209 @@ namespace ot {
...
+    Error(const Error &aError);
+    Error &operator=(const Error &aError);
+    Error(Error &&aError) noexcept;
+    Error &operator=(Error &&aError) noexcept;
+
+    /**
+     * Returns true if there is no error.
+     *
+     */
+    bool IsNone() const { return (mState == nullptr); }
```
simonlingoogle, line 186: nit:
```suggestion
    bool IsNone() const { return mState == nullptr; }
```

src/common/utils.hpp
```diff
@@ -36,13 +36,25 @@
...
 #include <commissioner/error.hpp>
 
-#define SuccessOrDie(aError)                   \
-    do                                         \
-    {                                          \
-        VerifyOrDie((aError) == Error::kNone); \
+#define ASSERT(aCondition)           \
+    do                               \
+    {                                \
+        bool condition = aCondition; \
```
simonlingoogle, line 47:
```suggestion
        bool condition = (aCondition); \
```

src/common/utils.hpp
```diff
@@ -36,13 +36,25 @@
...
+#include <assert.h>
+
 #include <commissioner/defines.hpp>
 #include <commissioner/error.hpp>
 
-#define SuccessOrDie(aError)                   \
-    do                                         \
-    {                                          \
-        VerifyOrDie((aError) == Error::kNone); \
+#define ASSERT(aCondition)           \
```
simonlingoogle, line 44: I wonder if it follows best practices to allow side-effects in `ASSERT`. It goes against conventions.

wgtdkp: I think we are not supposed to allowed side-effect in `ASSERT` unless we explicitly document it.
I did remove the comment about allowing side-effect os it is not supporting side-effect and if there are code includes side-effect in `ASSERT`, it should be treated as a bug.

simonlingoogle: aren’t this code allowing side effects?

include/commissioner/error.hpp
```diff
@@ -42,36 +43,209 @@ namespace ot {
...
+
+    /**
+     * The operation, transaction or message exchange was aborted.
+     */
+    kAborted = 14,
+
+    /**
+     * The commissioner is not in a valid state that the operation can be processed.
+     */
+
```
Irving-cl, line 141: remove the empty line?

include/commissioner/error.hpp
```diff
@@ -42,36 +43,209 @@ namespace ot {
...
+    kInvalidState = 15,
+
+    /**
+     * The operational has been rejected. For example, petition could be rejected
```
Irving-cl, line 145:
```suggestion
     * The operation was rejected. For example, petition could be rejected
```

include/commissioner/error.hpp
```diff
@@ -42,36 +43,209 @@ namespace ot {
...
+     * because of existing active commissioner.
+     */
+    kRejected = 16,
+
+    /**
+     * The error is out of the address space of OT Commissioner.
```
Irving-cl, line 151: What does this mean?

src/app/json_test.cpp
```diff
@@ -85,7 +85,7 @@ TEST_CASE("active-operational-dataset-encoding-decoding", "[json]")
 
         std::string              json = ActiveDatasetToJson(dataset);
         ActiveOperationalDataset dataset1;
-        REQUIRE(ActiveDatasetFromJson(dataset1, json) == Error::kNone);
+        REQUIRE(ActiveDatasetFromJson(dataset1, json).IsNone());
```
abtink, line 88: I feel this is not helping with readability of the code and that we are checking the there is no error.

I would suggest the following: In `Error` class we define an overload of `operator==(ErrorCode aCode)` which compares an `Error` with a `ErrorCode` enum value. This would allow original code (which in my opinion is more readable) to continue to work.

include/commissioner/error.hpp
```diff
@@ -42,36 +43,209 @@ namespace ot {
...
+    struct State
+    {
+        ErrorCode   mCode;
+        std::string mMessage;
+    };
+
+    // `None` error has a `NULL` mState. Otherwise, `mState` points to
+    // a `State` structure containing the error code and message(s).
+    // This is for performance, because a `None` error is the usual case.
+    std::unique_ptr<State> mState;
```
abtink, line 219: What is the reason behind the extra indirection and unique pointer use here?

Wouldn't using an `ErrorCode mCode` and `std::string mMessage` in `Error` be simpler and and avoid the dynamic allocation?

wgtdkp: In most situations, we will success and error will be `None`.  `std::string mMessage` typically occupies 32 bytes on 64 bits platform and the constructor or copy assignment will be called when the error is initialized or copied. The `unique_ptr` here is to make the error class smaller and avoid unnecessary copy.

FYI, our implementation here is inspired by [abseil/status.h](https://github.com/abseil/abseil-cpp/blob/master/absl/status/status.h) which has a similar imeplementation.

src/app/json_test.cpp
wgtdkp: Agree that `IsNone()` is confusing in some situations.

simonlingoogle: I used to suggest `IsSuccess()` for checking if the operation (error) was successful (none):

- `ActiveDatasetFromJson(dataset1, json).IsSuccess()` equally readable as (if not more readable than)  `ActiveDatasetFromJson(dataset1, json) == Error:kNone`
- `.IsSuccess()` is simply more consise than ` == Error::kNone`
- `.IsSuccess()` works better with editor's `autocomplete` feature (most of time we just need to type in `.`).

Thoughts? @wgtdkp @abtink @jwhui

wgtdkp: It will be a little weird when we defined a error:
```C++
Error error;
...
exit:
    if (error.IsSuccess())
    {
        ...
    }
```

`error is success` reads strange.

include/commissioner/error.hpp
wgtdkp: For  example, there are mbedtls errors that cannot fit into (convert to) OT Commissioner errors. In that case, this error code will be used and the error message will include the details of the mbedtls error.

src/app/json_test.cpp
abtink: I agree `error.IsSuccess()` may not be idle, but personally I think it is ok and would prefer `IsSuccess()` over `IsNone()` for statements like: `ActiveDatasetFromJson(dataset1, json).IsSuccess()`

include/commissioner/error.hpp
abtink: I guess this model is intended for use in code running on POSIX MCU/machines (i.e. not as resource constrained as for example OT core which can run on rather resource limited MCUs). So having an extra 32 bytes of memory per `Error` may be fine (particularly since `Error` instances are mostly temp/stack variables). 

On the other hand, I would expect the `std::string` copy constructor implementation to be efficient (specially for empty strings) and the advantage of using `ErrorCode mCode` and `std::string mMessage` is that the code would be simple and  avoids dynamic allocation and use of `unique_ptr` (which add their own overhead in terms of speed).

src/app/json_test.cpp
wgtdkp: @simonlingoogle @abtink Thanks for your advices! I think `operator==` works good in both situations. I have revised to use `==`, please review.

include/commissioner/error.hpp
wgtdkp: @abtink Agree that size and performance are not that important in context of an external Thread Commissioner. I removed the `unique_ptr`, please review.

simonlingoogle: But it-commissioner is indeed supposed to run on posix, so unique_ptr should be proper.

PR #49, opened 2020-03-28
[script] install cmake dependencies

wgtdkp:
this PR installs cmake dependency `scikit-build`, which is missing on some version of respberry Pi.

### DIFF
--- a/script/bootstrap.sh
+++ b/script/bootstrap.sh
@@ -81,6 +81,7 @@ if [ "$(uname)" = "Linux" ]; then
     ## Install newest CMake
     match_version "$(cmake --version | grep -E -o '[0-9].*')" "${MIN_CMAKE_VERSION}" || {
         pip install -U pip
+        pip install -U scikit-build
         pip install -U cmake
     }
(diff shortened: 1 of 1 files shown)


script/bootstrap.sh
```diff
@@ -81,6 +81,7 @@ if [ "$(uname)" = "Linux" ]; then
     ## Install newest CMake
     match_version "$(cmake --version | grep -E -o '[0-9].*')" "${MIN_CMAKE_VERSION}" || {
         pip install -U pip
+        pip install -U scikit-build
```
jwhui: Just curious, what error does this address? Is `scikit-build` a dependency specific to CMake or `ot-commissioner`?

wgtdkp: There may be [this issue](https://github.com/scikit-build/cmake-python-distributions/issues/86) on a raspberry Pi 3. Haven't ran into this before, but got the error when trying to bootstrap the commissioner on a newly flashed raspberry Pi 3 (Raspbian Stretch).

Issue #51, opened 2020-03-29
[mdns] remove mDNS from the commissioner library to CLI App.


wgtdkp:
This issue proposes remove current mDNS discovery from the commissioner library.

- The commissioner library is designed to be used for mobile Apps (Android & iOS), and they have better native mDNS/DNS service implementation. The functionalities of the commissioner library are not dependent on the mDNS service at all, so it may be better to remove it.
- We still provide border agent discovery functionality for the CLI App.

related issue: https://github.com/openthread/ot-commissioner/issues/31.

Issue #31, opened 2020-03-11
Support alternative mdns implementations


jwhui:
We support choosing between alternative mdns implementations in openthread/ot-br-posix using the ["--with-mdns" build option](https://github.com/openthread/ot-br-posix/blob/affbbdb34b2c6743b1e83a1fd5ef50cf6e93d6ac/configure.ac#L449).

wgtdkp:
We can even remove mdns from OT Commissioner library and let the application choose any mDNS implementation by providing a parser to parse the mDNS response into a structured [BorderAgent](https://github.com/openthread/ot-commissioner/blob/master/include/commissioner/border_agent.hpp) object.

Need to decide what is the best approach.

wgtdkp:
close this issue because mDNS is removed from the library to CLI App. An mobile App can use native mDNS implementation.

PR #53, opened 2020-03-29
[mdns] move mDNS out of the commissioner library

wgtdkp:
This PR moves discovery with mDNS out of the Commissioner library:

- It is better for the mobile App/OS to use native mDNS service rather than the one provided by this library.
- We still keep the `borderagent discover` command for the CLI App.

address issue: https://github.com/openthread/ot-commissioner/issues/51.
related issue: https://github.com/openthread/ot-commissioner/issues/31.


src/app/cli/interpreter.hpp
```diff
@@ -36,7 +36,8 @@
 
 #include <map>
 
-#include "commissioner_app.hpp"
+#include <border_agent.hpp>
+#include <commissioner_app.hpp>
```
bukepo, line 40:
```suggestion
#include "app/border_agent.hpp"
#include "app/commissioner_app.hpp"
```

wgtdkp: Thanks, will address this for all headers in separate PR.

PR #59, opened 2020-04-08
[socket] add missing header guard

wgtdkp:


src/library/socket.hpp
```diff
@@ -31,6 +31,10 @@
  *   The file includes definitions of Socket.
  */
 
+
+#ifndef SOCKET_HPP_
```
bukepo, line 35:
```suggestion
#ifndef OT_COMM_LIBRARY_SOCKET_HPP_
```

wgtdkp: Since the header file is only accessible inside `src/library`, do we need the prefix here?

wgtdkp: maybe `LIBRARY_*`

gocarlos: I'd use `#pragma once` as this makes sure, no crashes with other macros and is supported by all compilers available for  CPP

gocarlos: I'd use `#pragma once` as this makes sure, no crashes with other macros and is supported by all compilers available for  CPP

wgtdkp: @gocarlos `#gragma once` doesn't seem to be a C++ standard.

gocarlos: No, it is not, pragmas are vendor specific stuff.
Though is supported by all compilers this library supports and is used be a ton of c/cpp libs...
I don't have a strong opinion on this, i’m just pragmatic and saw many errors with other include guards

wgtdkp: okay, I would prefer to keep current header guards to be consistent with openthread projects.

bukepo: https://google.github.io/styleguide/cppguide.html#The__define_Guard

wgtdkp: @bukepo Will do that in separate PR.

wgtdkp:
@gocarlos @bukepo more comments?

Issue #63, opened 2020-04-08
update cn-cbor


gocarlos:
after https://github.com/cabo/cn-cbor/issues/60 the author decided to archive the project, therefore the `complete` branch in https://github.com/jimsch/cn-cbor was merged to master...

we could therefore update to track master in the submodule here

PR #64, opened 2020-04-08
[cn-cbor] upgrade and track master branch

gocarlos:
closes https://github.com/openthread/ot-commissioner/issues/63


gocarlos:
> Thanks @gocarlos ! LGTM 👍
> 
> By the way, I saw you contribution to the cn-cbor project to make it work with modern CMake. Do you think we still need the cn-cbor wrapper CMakeList file here?

No, i think, we could add cn-cbor directly with `add_subdirectory` and use the target `cn-cbor::cn-cbor`

wgtdkp:
> No, i think, we could add cn-cbor directly with add_subdirectory and use the target cn-cbor::cn-cbor

Thanks, will do the update.

PR #66, opened 2020-04-10
[build] add native cn-cbor library

wgtdkp:
This PR adds native cn-cbor library instead of the one crafted.


wgtdkp:
@gocarlos Could you help review this PR? Thanks!

PR #68, opened 2020-04-14
Fix incomplete joiner command help

deimi:


deimi:
@googlebot I signed it!

PR #73, opened 2020-04-16
[provisioning-url] move matching to application layer

deimi:
This way it would be possible to let every device join, no matter what kind of application layer provisioning is required later.

### DIFF
--- a/src/app/commissioner_app.cpp
+++ b/src/app/commissioner_app.cpp
@@ -1318,23 +1318,18 @@ bool CommissionerApp::HandleCommissioning(const JoinerInfo & aJoinerInfo,
     (void)aVendorStackVersion;
     (void)aVendorData;
 
-    constexpr auto kJoinerNotAccepted = false;
-    constexpr auto kJoinerAccepted    = true;
+    bool accepted = false;
 
     auto configuredJoinerInfo = GetJoinerInfo(aJoinerInfo.mType, Commissioner::ComputeJoinerId(aJoinerInfo.mEui64));
-    if (configuredJoinerInfo == nullptr)
-    {
-        // TODO(wgtdkp): logging
-        return kJoinerNotAccepted;
-    }
 
-    if (aProvisioningUrl != configuredJoinerInfo->mProvisioningUrl)
-    {
-        // TODO(wgtdkp): logging
-        return kJoinerNotAccepted;
-    }
+    // TODO(deimi): logging
+    VerifyOrExit(configuredJoinerInfo != nullptr, accepted = false);
+    VerifyOrExit(aProvisioningUrl == configuredJoinerInfo->mProvisioningUrl, accepted = false);
 
-    return kJoinerAccepted;
+    accepted = true;
+
+exit:
+    return accepted;
 }
 
 } // namespace commissioner
--- a/src/app/commissioner_app.cpp
+++ b/src/app/commissioner_app.cpp
@@ -45,29 +45,6 @@ namespace ot {
 
 namespace commissioner {
 
-/**
- * The default commissioning handler that always accepts any joiner.
- *
- */
-static bool DefaultCommissioningHandler(const JoinerInfo & aJoinerInfo,
-                                        const std::string &aVendorName,
-                                        const std::string &aVendorModel,
-                                        const std::string &aVendorSwVersion,
-                                        const ByteArray &  aVendorStackVersion,
-                                        const std::string &aProvisioningUrl,
-                                        const ByteArray &  aVendorData)
-{
-    (void)aJoinerInfo;
-    (void)aVendorName;
-    (void)aVendorModel;
-    (void)aVendorSwVersion;
-    (void)aVendorStackVersion;
-    (void)aProvisioningUrl;
-    (void)aVendorData;
-
-    return true;
-}
-
 std::shared_ptr<CommissionerApp> CommissionerApp::Create(const std::string &aConfigFile)
 {
     Error     error = Error::kNone;
@@ -107,7 +84,13 @@ Error CommissionerApp::Init(const AppConfig &aAppConfig)
         [this](JoinerType aType, const ByteArray &aJoinerId) { return GetJoinerInfo(aType, aJoinerId); });
 
     // This is the default behavior of OpenThread on-Mesh Commissioner.
-    mCommissioner->SetCommissioningHandler(DefaultCommissioningHandler);
+    mCommissioner->SetCommissioningHandler([this](const JoinerInfo &aJoinerInfo, const std::string &aVendorName,
+                                                  const std::string &aVendorModel, const std::string &aVendorSwVersion,
+                                                  const ByteArray &  aVendorStackVersion,
+                                                  const std::string &aProvisioningUrl, const ByteArray &aVendorData) {
+        return HandleCommissioning(aJoinerInfo, aVendorName, aVendorModel, aVendorSwVersion, aVendorStackVersion,
+                                   aProvisioningUrl, aVendorData);
+    });
 
 exit:
     return error;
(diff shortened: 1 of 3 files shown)


deimi:
As far as I understand the code, the check for a valid provisioning url could also be completely deleted, because the provisioning url is nevertheless checked by the CommissioningHandler in the application code. So it makes no sense for me that the library is checking the url as well.

wgtdkp:
Thanks! @deimi 
If the provisioning URL in `mJoinerInfo` is empty, it means the commissioner doesn't support vendor-specific provisioning for this device. So, if a joiner requires vendor-specific provisioning while the commissioner cannot handle it, the joiner should be rejected, which is defined behavior by Thread 1.1 specification.

I agree that it seems not necessary for the commissioner library to check the provisioning URL itself before delegating to application, while we made it there to just enforce that a commissioner app follows the specification. I am fine with moving the provisioning URL matching from commissioner library to application layer, but to compliant to Thread specification, we still should not accept a joiner if the provisioning URLs don't match at there. Thoughts?

deimi:
@wgtdkp Ah ok, I understand your points for pre matching in the library.

Yes that would be my suggestion too. Remove the url matching from the library and let only decide the application via the CommissioningHandler if the provisioning url is accepted. The library should only check the return value of the CommissioningHandler.
Nevertheless the library should still check if the joiner requests application layer provisioning, but the application has not set a CommissioningHandler. This is already in the code.

This way the library can also cover the use case where an application is capable of handling several different provisioning urls.

wgtdkp:
@deimi 
> The library should only check the return value of the CommissioningHandler.

This totally works for me. Could you make the change to that instead?

> This way the library can also cover the use case where an application is capable of handling several different provisioning urls.

Please notice that the provisioning URL is per joiner. The application can specify different provisioning URL for each joiner, but don't support multiple provisioning URL of a single joiner device (Is there real use case of this?).

deimi:
@wgtdkp 
> This totally works for me. Could you make the change to that instead?
> 
:+1: 

> Please notice that the provisioning URL is per joiner. The application can specify different provisioning URL for each joiner, but don't support multiple provisioning URL of a single joiner device (Is there real use case of this?).

Especially during application development this could be handy. That's the reason why I stumbled across this, because I just used the cli app with ```joiner enableall meshcop XYZ``` :sweat_smile: 
Another situation which comes in my mind is that a joiner could run with different firmwares with different provisioning url configured. After this merge the matching lies in the hands of the application and can therefore handle different urls for the same joiner eui64.

wgtdkp:
@deimi 
> joiner enableall meshcop XYZ

Enable a joiner without a provisioning URL is legal (but the joiner password is illegal), this means the commissioner doesn't support vendor-specific provisioning for this joiner.


> Another situation which comes in my mind is that a joiner could run with different firmwares with different provisioning url configured. After this merge the matching lies in the hands of the application and can therefore handle different urls for the same joiner eui64.

For real product, I think the provisioning URL is not going to be changed but set when created from the factory...

deimi:
@wgtdkp 

> For real product, I think the provisioning URL is not going to be changed but set when created from the factory...

Probably not, I was just thinking about a bug during update or something else.
Nevertheless now this would be up to the application layer to handle this.

src/library/commissioning_session.cpp
```diff
@@ -181,22 +181,22 @@ void CommissioningSession::HandleJoinFin(const coap::Request &aJoinFin)
...
         VerifyOrExit(vendorDataTlv != nullptr, error = Error::kNotFound);
         VerifyOrExit(vendorDataTlv->IsValid(), error = Error::kBadFormat);
-        VerifyOrExit(provisioningUrlTlv->GetValueAsString() == mJoinerInfo.mProvisioningUrl, error = Error::kReject);
 
         provisioningUrl = provisioningUrlTlv->GetValueAsString();
         vendorData      = vendorDataTlv->GetValue();
     }
 
+    LOG_INFO("received JOIN_FIN.req: vendorName={}, vendorModel={}, vendorSWVversion={}, "
+             "vendorStackVersion={}, provisioningUrl={}, vendorData={:#04x}",
```
wgtdkp, line 195: vendorData is a `ByteArray` but not an integer, maybe print the hex string?

src/library/commissioning_session.cpp
```diff
@@ -181,22 +181,22 @@ void CommissioningSession::HandleJoinFin(const coap::Request &aJoinFin)
...
-             "vendorStackVersion={}",
-             vendorNameTlv->GetValueAsString(), vendorModelTlv->GetValueAsString(),
-             vendorSwVersionTlv->GetValueAsString(), utils::Hex(vendorStackVersionTlv->GetValue()));
-
     if (auto provisioningUrlTlv = tlvSet[tlv::Type::kProvisioningURL])
     {
         auto vendorDataTlv = tlvSet[tlv::Type::kVendorData];
         VerifyOrExit(vendorDataTlv != nullptr, error = Error::kNotFound);
         VerifyOrExit(vendorDataTlv->IsValid(), error = Error::kBadFormat);
-        VerifyOrExit(provisioningUrlTlv->GetValueAsString() == mJoinerInfo.mProvisioningUrl, error = Error::kReject);
```
wgtdkp, line 194: We need to at least add this to the `CommisioningHandler` in `CommissionerApp`:
https://github.com/openthread/ot-commissioner/blob/4f63a0f9bbfe9db2106ea8decf784a4399ef67a2/src/app/commissioner_app.cpp#L52

You may need to change the static default handle to a member function of class `CommissionerApp` now.

src/library/commissioning_session.cpp
deimi: `{:#04x}` will print a hex string

wgtdkp: `fmt::join()` makes sense. Should we use `{:#02x}` instead? By the way, we have `utils::Hex()` to cast `ByteArray` to a continuous hex string (e.g. `utils::Hex(vendorStackVersionTlv->GetValue())`). Your format looks nice, but it will be inconsistent with the rest. Would mind to use the `utils::Hex()`?

We may define `utils::PrettyHex()` to add delimiters to the hex string when used in logging messages (but not necessary in this PR). Thoughts?

deimi: Oh, I missed utils::Hex(). I will change it to stay consistent.

src/library/commissioning_session.cpp
deimi: Changed the CommissionerApp to only accept configured provisioning urls

deimi:
@wgtdkp please have a look

src/app/commissioner_app.cpp
```diff
@@ -1321,6 +1304,39 @@ const JoinerInfo *CommissionerApp::GetJoinerInfo(JoinerType aType, const ByteArr
...
+    (void)aVendorData;
+
+    constexpr auto kJoinerNotAccepted = false;
+    constexpr auto kJoinerAccepted    = true;
+
+    auto configuredJoinerInfo = GetJoinerInfo(aJoinerInfo.mType, Commissioner::ComputeJoinerId(aJoinerInfo.mEui64));
+    if (configuredJoinerInfo == nullptr)
+    {
+        // TODO(wgtdkp): logging
+        return kJoinerNotAccepted;
```
wgtdkp, line 1328: @deimi `OpenThread` doesn't encourage multiple `return` in a function, could you consider using `VerifyOrExit(...)`? for example, you can change this to
```c++
VerifyOrExit(configuredJoinerInfo != nullptr, accepted = false);
```

src/app/commissioner_app.cpp
```diff
@@ -1321,6 +1304,39 @@ const JoinerInfo *CommissionerApp::GetJoinerInfo(JoinerType aType, const ByteArr
...
+    (void)aVendorStackVersion;
+    (void)aVendorData;
+
+    constexpr auto kJoinerNotAccepted = false;
+    constexpr auto kJoinerAccepted    = true;
+
+    auto configuredJoinerInfo = GetJoinerInfo(aJoinerInfo.mType, Commissioner::ComputeJoinerId(aJoinerInfo.mEui64));
+    if (configuredJoinerInfo == nullptr)
+    {
+        // TODO(wgtdkp): logging
```
wgtdkp, line 1327: Maybe have your ID here?

src/library/commissioning_session.cpp
```diff
@@ -209,6 +209,7 @@ void CommissioningSession::HandleJoinFin(const coap::Request &aJoinFin)
         // Accepts a joiner if requirement on vendor-specific provisioning.
         accepted = provisioningUrl.empty();
     }
+    VerifyOrExit(accepted == true, error = Error::kReject);
```
wgtdkp, line 212:
```suggestion
    VerifyOrExit(accepted, error = Error::kReject);
```

src/app/commissioner_app.cpp
deimi: I will, but just to mention it, there are several multiple return functions in `commissioner_app.cpp`

wgtdkp: oh, could you help point out them? I am appreciate if you even like to make a PR to fix it.

deimi: e.g. `GetJoinerInfo`

wgtdkp: Thanks, I will fix those multiple returns.

PR #74, opened 2020-04-17
[config]  refine commissioner configuration and logger

wgtdkp:
This PR refine the configuration and logger:
1. Defines a new `Logger` abstract class to easily customize an application-provided logger.
2. Throw exception when parsing `Network Data` and `Config` from json file failed and convert it to the commissioner `Error`. This give the chance to return user-friendly error message for the configuration file (or Network Data file) when https://github.com/openthread/ot-commissioner/pull/45 is merged.
3. Separate file utility to `file_util.cpp`.
4. Separate file logging to `file_logger.cpp`.
5. Initialize `CommissionerApp` directly with `Config` and remove the intermediate `AppConfig`. This enable unit tests for `CommissionerApp` (coming with separate PR).

### DIFF
--- a/src/app/file_util.cpp
+++ b/src/app/file_util.cpp
@@ -1,5 +1,5 @@
 /*
- *    Copyright (c) 2019, The OpenThread Authors.
+ *    Copyright (c) 2020, The OpenThread Authors.
  *    All rights reserved.
  *
  *    Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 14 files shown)


src/app/file_util.cpp
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    return error;
+}
+
+Error ReadFile(std::string &aData, const std::string &aFilename)
+{
+    Error error = Error::kNone;
+
+    FILE *f = fopen(aFilename.c_str(), "r");
```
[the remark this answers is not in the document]
wgtdkp: There are two main reasons:
1. I has been writing `std::ofstream file(aFilename, ...)` to open a file, but forget to handle the exception from the constructor. We can also use the `open()` `close` methods, but:
2. With the c way, `errno` will be set when failed, and we get a chance to return a user-friendly error message by `perror()` after adding error message to the `Error` class (https://github.com/openthread/ot-commissioner/pull/45). While I don't find if C++ `std::fstream` can make those error message accessible (http://www.cplusplus.com/reference/fstream/fstream/).

src/app/json.cpp
```diff
@@ -28,18 +28,61 @@
...
+} // namespace commissioner
+
+} // namespace ot
+
+#define SuccessOrThrow(aError)                                                                  \
+    do                                                                                          \
+    {                                                                                           \
+        if (aError != ::ot::commissioner::Error::kNone)                                         \
+        {                                                                                       \
+            throw ::ot::commissioner::JsonException(::ot::commissioner::ErrorToString(aError)); \
```
jwhui, line 82: Google C++ style guide strongly recommends against the use of exceptions. Why do we use them here?

wgtdkp: Because the [json](https://github.com/nlohmann/json) library uses exception exclusively. The macro `SuccessOrThrow` is only defined in this `json.cpp` file and only called in those converters (`from_json()` & `to_json()`) called by the json library. see also the doc string of the `JsonException` class above.
 
All APIs in `json.hpp` are guaranteed never throw a exception.

jwhui: Makes sense. Thanks for the clarification!

PR #82, opened 2020-04-24
[pretty] add markdown support

jwhui:


script/make-pretty
```diff
@@ -71,17 +78,37 @@ do_clang_check()
...
 }
 
+do_markdown_format()
+{
+    echo -e '======================'
+    echo -e '     format markdown'
+    echo -e '======================'
+
+    git ls-files "${OT_MARKDOWN_SOURCES[@]}" | grep -v -E "^($(echo "${OT_EXCLUDE_DIRS[@]}" | tr ' ' '|'))" \
+        | xargs -n10 -P"${OT_BUILD_JOBS}" npx prettier@2.0.4 --write
```
simonlingoogle, line 92: need install `npx` in `bootstrap.sh`?

jwhui:
> BTW, how about formatting CMake files with [cmake-format](https://github.com/cheshirekow/cmake_format)? Since there are practices to [treat CMake file as code](https://gist.github.com/mbinna/c61dbb39bca0e4fb7d1f73b0d66a4fd1#treat-cmake-code-like-production-code) :)

Sounds good to me! I can help take a look.

script/make-pretty
jwhui: I can address this in a separate PR.

PR #83, opened 2020-04-24
[event-loop] break event loop from inside

wgtdkp:
# Problem
We were directly calling [event_base_loopbreak](http://www.wangafu.net/~nickm/libevent-2.0/doxygen/html/event_8h.html#a07a7599e478e4031fa8cf52e26d8aa1e) to stop the event loop running in background thread from user thread. It has a problem that if `event_base_loopbreak` is called even before the event loop has been started (`event_base_loop` has not been called), the event loop will not be breaked (because it has not been started yet). This will result in dead loop when we stop immediately after starting the commissioner.
```text
T     user thread                                      event loop thread
----------------------------------------------------------------------------------------
t0    commissioner->Start():start event loop thread
----------------------------------------------------------------------------------------             
t1    commissioner->Stop(): try to break event loop    (no effect)
---------------------------------------------------------------------------------------- 
t2                                                     run event loop
---------------------------------------------------------------------------------------- 
t3    join event loop thread and wait forever.
---------------------------------------------------------------------------------------- 
...
```

# Solve
This PR fixes this issue by sending `event_base_loopbreak` to the event loop and breaks it from inside the event loop. This makes sure that the event loop has been started when `event_base_loopbreak` is actually executed.
```text
T     user thread                                         event loop thread
----------------------------------------------------------------------------------------
t0    commissioner->Start():start event loop thread
----------------------------------------------------------------------------------------              
t1    commissioner->Stop(): send event-loop-break       
      event and wait for it being processed.
---------------------------------------------------------------------------------------- 
t2    waiting...                                          run event loop: process event-loop-break event
---------------------------------------------------------------------------------------- 
t3                                                        event loop is stopped and event loop thread exits.
----------------------------------------------------------------------------------------
t4    join event loop thread
----------------------------------------------------------------------------------------
t5    commissioner->Stop() exit
...
```

This PR also replaces `std::shared_ptr<std::thread>` with `std::thread` and use [`std::thread::joinable`](https://en.cppreference.com/w/cpp/thread/thread/joinable) to uniquely identifies whether the event loop thread is actively running and joinable.

### DIFF
--- a/src/library/commissioner_safe.cpp
+++ b/src/library/commissioner_safe.cpp
@@ -61,7 +61,6 @@ std::shared_ptr<Commissioner> Commissioner::Create(const Config &aConfig, struct
 
 CommissionerSafe::CommissionerSafe()
     : mImpl(mEventBase.Get())
-    , mEventThread(nullptr)
 {
 }
 
@@ -96,9 +95,9 @@ Error CommissionerSafe::Start()
 {
     Error error = Error::kNone;
 
-    VerifyOrExit(mEventThread == nullptr, error = Error::kAlready);
+    VerifyOrExit(!mEventThread.joinable(), error = Error::kAlready);
 
-    mEventThread = std::make_shared<std::thread>([this]() { IgnoreError(mImpl.Start()); });
+    mEventThread = std::thread([this]() { IgnoreError(mImpl.Start()); });
 
 exit:
     return error;
@@ -108,6 +107,8 @@ exit:
 void CommissionerSafe::Stop()
 {
     std::promise<void> pro;
+
+    VerifyOrExit(mEventThread.joinable());
 
     // Send `Stop` to the event loop to break it from inside.
     // This makes sure the event loop has been started when we
@@ -119,11 +120,9 @@ void CommissionerSafe::Stop()
 
     pro.get_future().wait();
 
-    if (mEventThread && mEventThread->joinable())
-    {
-        mEventThread->join();
-        mEventThread = nullptr;
-    }
+    mEventThread.join();
+
+exit:
     return;
 }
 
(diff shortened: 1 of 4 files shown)


src/library/commissioner_safe.cpp
```diff
@@ -96,9 +95,9 @@ Error CommissionerSafe::Start()
 {
     Error error = Error::kNone;
 
-    VerifyOrExit(mEventThread == nullptr, error = Error::kAlready);
+    VerifyOrExit(!mEventThread.joinable(), error = Error::kAlready);
```
simonlingoogle, line 98: Could two threads calls `Start()` at the same time and `mEventThread.joinable()` returns False for both thread. Both threads will then try to create `mEventThread`?

I suspect all methods of `CommissionerSafe` need to be thread safe?

src/library/commissioner_safe.cpp
```diff
@@ -107,13 +106,23 @@ Error CommissionerSafe::Start()
...
 {
-    mImpl.Stop();
+    std::promise<void> pro;
 
-    if (mEventThread && mEventThread->joinable())
-    {
-        mEventThread->join();
-        mEventThread = nullptr;
-    }
+    VerifyOrExit(mEventThread.joinable());
```
simonlingoogle, line 111: Multiple threads can pass `VerifyOrExit` at the same time?

src/library/commissioner_safe.cpp
wgtdkp: Yes it is not thread safe. I think the thread safety promise of this class is much. I updated the [doc string of class CommissionerSafe](). Please check if it makes sens for you.

simonlingoogle: Sure, let's keep it simple.

src/library/commissioner_safe.cpp
wgtdkp: it is not safe to be called from multiple user threads. mark this as resolved becuase we lowered the thread safety of this `CommissionerSafe` to be only safe between one single user thread and the event loop thread.

PR #86, opened 2020-04-28
[app] clear network data when stopped

wgtdkp:
This PR clears data associated with the connected Thread Network when the commissioner is stopped (disconnected from the Thread Network).

### DIFF
--- a/src/app/commissioner_app.cpp
+++ b/src/app/commissioner_app.cpp
@@ -112,6 +112,14 @@ exit:
 void CommissionerApp::Stop()
 {
     IgnoreError(mCommissioner->Resign());
+
+    mJoiners.clear();
+    mPanIdConflicts.clear();
+    mEnergyReports.clear();
+    mActiveDataset  = ActiveOperationalDataset();
+    mPendingDataset = PendingOperationalDataset();
+    mCommDataset    = MakeDefaultCommissionerDataset();
+    mBbrDataset     = BbrDataset();
 }
 
 void CommissionerApp::AbortRequests()
--- a/src/app/commissioner_app.hpp
+++ b/src/app/commissioner_app.hpp
@@ -234,8 +234,6 @@ private:
 
     ByteArray mSignedToken;
 
-    std::list<BorderAgent> mBorderAgents;
-
 private:
     void HandlePanIdConflict(const std::string *aPeerAddr,
                              const ChannelMask *aChannelMask,
--- a/src/app/commissioner_app.hpp
+++ b/src/app/commissioner_app.hpp
@@ -48,6 +48,7 @@
 #include <commissioner/commissioner.hpp>
 #include <commissioner/network_data.hpp>
 
+#include "app/border_agent.hpp"
 #include "common/address.hpp"
 
 namespace ot {
@@ -231,9 +232,9 @@ private:
 
     std::shared_ptr<Commissioner> mCommissioner;
 
-    std::map<JoinerKey, JoinerInfo> mJoiners;
-
     ByteArray mSignedToken;
+
+    std::list<BorderAgent> mBorderAgents;
 
 private:
     void HandlePanIdConflict(const std::string *aPeerAddr,
@@ -244,9 +245,6 @@ private:
                             const ChannelMask *aChannelMask,
                             const ByteArray *  aEnergyList,
                             Error              aError);
-
-    std::map<uint16_t, ChannelMask> mPanIdConflicts;
-    EnergyReportMap                 mEnergyReports;
 
     void HandleDatasetChanged(Error error);
 
@@ -259,12 +257,15 @@ private:
                              const ByteArray &  aVendorData);
 
     /*
-     * Below are network data associated to the connected Thread network.
+     * Below are data associated with the connected Thread Network.
      */
-    ActiveOperationalDataset  mActiveDataset;
-    PendingOperationalDataset mPendingDataset;
-    CommissionerDataset       mCommDataset;
-    BbrDataset                mBbrDataset;
+    std::map<JoinerKey, JoinerInfo> mJoiners;
+    std::map<uint16_t, ChannelMask> mPanIdConflicts;
+    EnergyReportMap                 mEnergyReports;
+    ActiveOperationalDataset        mActiveDataset;
+    PendingOperationalDataset       mPendingDataset;
+    CommissionerDataset             mCommDataset;
+    BbrDataset                      mBbrDataset;
 };
 
 } // namespace commissioner


src/app/commissioner_app.cpp
```diff
@@ -112,6 +112,14 @@ Error CommissionerApp::Start(std::string &      aExistingCommissionerId,
 void CommissionerApp::Stop()
```
simonlingoogle, line 112: Not directly related to this PR, but I am wondering why do we need a `CommissionerApp` to wrap `CommissionerSafe` ? Which are users supposed to use in their app?

src/app/commissioner_app.cpp
```diff
@@ -112,6 +112,14 @@ Error CommissionerApp::Start(std::string &      aExistingCommissionerId,
...
 {
     IgnoreError(mCommissioner->Resign());
+
+    mJoiners.clear();
```
simonlingoogle, line 116: Are we allowing `CommissionerApp` to be reused for multiple times (`Start`, `Stop` for multiple times) ? 
One design choice is to not allow a stopped `CommissionerApp` to be restarted. In this way we could simplify state management for `CommissionerApp`?

src/app/commissioner_app.cpp
wgtdkp: commissioner library (targets in `src/library`) is the one will be used for developing Apps (please notice that `CommissionerApp` is under `src/app`).

Q: Why do we need `CommissionerApp`?

A: The commissioner library is stateless, it doesn't maintains network datas of the connected Thread Network . `CommissionerApp` maintains network data and events from the commissioner library. The reason why commissioner library doesn't maintain network data is that the network data doesn't belong to the commissioner library but Apps and it is better for mobile Apps to maintain network data by themselves.  `CommissionerApp` also provides more user-friendly  methods that can be directly mapped to CLI commands so we can keep cli/interpreter.cpp just command parsing stuffs.

src/app/commissioner_app.cpp
wgtdkp: Yes, we allow `CommissionerApp` to be restarted for multiple times. Currently, `CommissionerApp` is for just commissioner CLI. and methods `Start()`, `Stop` are mapped to CLI command `start ...`, `stop`.

src/app/commissioner_app.cpp
simonlingoogle: >  `CommissionerApp` also provides more user-friendly methods that can be directly mapped to CLI commands so we can keep cli/interpreter.cpp just command parsing stuffs.

Even though `CommissionerSafe` is supposed to be used by developers, from what you said, developers might just prefer to use `CommissionerApp`?

wgtdkp: When saying "user-firendly", I mean the user of commissioner CLI but not developer. 

Why we don't provide `CommissionerApp`'s APIs in commissioner library:
1. methods of `CommissionerApp` are simple wrap of each API of the commissioner library, and they are the synchronous version. For a Commissioner App developer, they would want to decide using sync or async APIs by themselves. For CLI, we did choose synchronous APIs.
2. It is not flexible. methods of `CommissionerApp` are simple wrap of each API of the commissioner library. There are too APIs for each TLV in each network dataset. Take BBR dataset for example, we have five APIs in`CommissionerApp`:
    ```c++
    Error GetTriHostname(std::string &aHostname) const;
    Error SetTriHostname(const std::string &aHostname);
    Error GetRegistrarHostname(std::string &aHostname) const;
    Error SetRegistrarHostname(const std::string &aHostname);
    // Registrar IPv6 Address is read-only.
    Error GetRegistrarIpv6Addr(std::string &aIpv6Addr) const;
    ```
    If we provide those in the commissioner library, we will need to provide 10 APIs, because we need to provide asynchronous version too. Even worse, if Thread spec is going to add a new TLV in a dataset, we need to add 4 more APIs to the commissioner library, which I think is what a commissioner library should avoid. But with our current design, we only need to add a new member to the dataset structure.
3. The commissioner library doesn't own/maintain network data, but provide APIs to query or change network data. The network data belongs to the network maintainer, Commissioner App. For Commissioner App developers, they may decide whether use their local cache of the network data or send `MGMT_XXX` if the cache is expired. In our CLI App, we use different strategies for different datasets. If we put network data into commissioner library, we need to double the number of the API (one for querying from local cache and another for querying from connected Thread network) or we need to add an additional parameter to each API to indicate whether should this API send `MGMT_XXX` message. It complicates the commissioner library and I think we should avoid it.

For overall understanding of the commissioner library and `CommissionerApp`, please take the commissioner library as `a library provides APIs to handle meshcop commissioning session and send MGMT messages for App developers`, and the `CommissionerApp` as `a Commissioner CLI App that maintains network data, user data and provides an interactive interface for a human user`.

simonlingoogle: Seems to me that maintaining network data is direct related to the job of a commissioner. 

Do you think it's beneficial in the future that we provide a library class similar to `CommissionerApp`, maybe named `ThreadNetworkManager`, for managing the Thread network? 

If I am the developer who is developing an APP for managing Thread network, I would just love using `CommissionerApp::SetNetworkName` without worrying about how to manipulate `Active Datasets`.
https://github.com/openthread/ot-commissioner/blob/715151f309469c73620c4aa0f7a8083db9766b31/src/app/commissioner_app.cpp#L627-L644

wgtdkp: > Seems to me that maintaining network data is direct related to the job of a commissioner.

Yes, but it is the part of the commissioner library.

> Do you think it's beneficial in the future that we provide a library class similar to CommissionerApp, maybe named ThreadNetworkManager, for managing the Thread network?

The `ThreadNetworkManager` will have more than 100 APIs. take `Network Name zTLV` for example, we need to provide `AsyncSetNetworkName`, `SyncSetNetworkName`, `AsyncGetNetworkName`, `SyncGetNetworkName` and `GetNetworkNameCached` ...  `CommissionerApp` is facing huan user, I think it is the App developer to define what the end user interface is.

> If I am the developer who is developing an APP for managing Thread network, I would just love using CommissionerApp::SetNetworkName without worrying about how to manipulate Active Datasets.

This may be convenient, but not flexible for developing a commissioner library as I said. If you you are going to develop an App for managing Thread network, I believe you have already understood `Active Dataset` well and know how to manipulate it (it is really not complicate at all if you know about `Active Dataset`), right?

simonlingoogle: > > Seems to me that maintaining network data is direct related to the job of a commissioner.
> 
> Yes, but it is not part of the commissioner library.

I would argue that `Network Management` is directly related to `Commissioner` as Spec states:
```
8.7.2.3 MGMT_SET.req – Set Management Data Request
MGMT_SET.req is sent from the Commissioner to the Leader. The response to this, the
29 command MGMT_SET.rsp, is sent from the Leader to the Commissioner. 
```
And we actually have already add these `Network Management` code (i.e. `CommissionerApp`) in `ot-commissioner` project. It is just one step away for us to transfer `CommissionerApp` to `ThreadNetworkManager`. 

> The `ThreadNetworkManager` will have more than 100 APIs. take `Network Name TLV` for example, we need to provide `AsyncSetNetworkName`, `SyncSetNetworkName`, `AsyncGetNetworkName`, `SyncGetNetworkName` and `GetNetworkNameCached` ... `CommissionerApp` is facing CLI user, I think it is the App developer to define what the end user interface is.

I think we can have the design choice to provide only one API for retrieving `Network Name`, which will be called `GetNetworkName`. A network manager might just like to wait for the result of `GetNetworkName`, so it can just be **synchronous**. We also don't need to provide API like `GetNetworkNameCached` because synchronizing datasets with Leader can be the implementation detail of the Network Manager which users don't need to worry about.  

> > If I am the developer who is developing an APP for managing Thread network, I would just love using CommissionerApp::SetNetworkName without worrying about how to manipulate Active Datasets.
> 
> This may be convenient, but not flexible for developing a commissioner library as I said. If you you are going to develop an App for managing Thread network, I believe you have already understood `Active Dataset` well and know how to manipulate it (it is really not complicate at all if you know about `Active Dataset`), right?

Not necessarily? If we provide `NetworkManager` with well defined easy-to-use APIs, the developers without full knowledge of `Commissioning` can also work well with the `NetworkManager.

wgtdkp: I just didn’t let the commissioner **library** directly own the thread network data, but not think that thread network management is not part of a commissioner’s functionality. Otherwise, why did we provide those APIs to get/set network datasets?

PR #87, opened 2020-04-28
 [tests] fix unit tests not found by catch2

wgtdkp:
This PR includes mainly two changes:
1. Fix catch2 cannot find testcases in linked libraries by defining `commissioner-common-test` and `commissioner-app-test` as `OBJECT` targets.
2. cleanup empty unittest files.


wgtdkp:
@bukepo suggest review the change with "Hide whitespace changes".

wgtdkp:
> I think you may consider using ctest.

Thanks for the suggestion, yes, it may be a good improvement to do.

PR #88, opened 2020-04-28
[logging] make FileLogger thread safe

wgtdkp:
This PR:
1. Makes concurrent call to `FileLogger::Log(...)` safe.
2. Refactors `FileLogger` to return `Error` when construction failed.

### DIFF
--- a/src/app/json.cpp
+++ b/src/app/json.cpp
@@ -75,13 +75,13 @@ private:
 
 } // namespace ot
 
-#define SuccessOrThrow(aError)                                  \
-    do                                                          \
-    {                                                           \
-        if (aError != ::ot::commissioner::Error::kNone)         \
-        {                                                       \
-            throw ::ot::commissioner::JsonException(aError);    \
-        }                                                       \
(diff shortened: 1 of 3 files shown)


src/app/json.cpp
```diff
@@ -213,8 +213,10 @@ static void from_json(const Json &aJson, Config &aConfig)
 
     if (aJson.contains("LogFile"))
     {
-        auto logger     = std::make_shared<FileLogger>(aJson["LogFile"], logLevel);
-        aConfig.mLogger = std::dynamic_pointer_cast<Logger>(logger);
+        std::shared_ptr<FileLogger> logger;
+
+        SuccessOrThrow(FileLogger::Create(logger, aJson["LogFile"], logLevel));
```
simonlingoogle, line 218: `SuccessOrThrow` should only be called for json related code?

wgtdkp: Yes, will refactor this.

simonlingoogle: Fine to use exception only in this file.

PR #94, opened 2020-05-02
[third-party] use libevent 2.1.11-stable

wgtdkp:
This PR set the version of libevent to [2.1.11-stable](https://github.com/libevent/libevent/tree/release-2.1.11-stable).

The latest libevent master introduced pthread features ([pthread_mutexattr_setprotocol](https://www.gnu.org/software/gnulib/manual/html_node/pthread_005fmutexattr_005fsetprotocol.html)) that are not supported on Android 5 (API level = 21).

### DIFF
--- a/src/library/CMakeLists.txt
+++ b/src/library/CMakeLists.txt
@@ -86,6 +86,7 @@ target_link_libraries(commissioner
         fmt::fmt
         event_core
         event_pthreads
+        pthread
         commissioner-common
 )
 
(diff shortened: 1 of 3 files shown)


src/library/CMakeLists.txt
```diff
@@ -86,6 +86,7 @@ target_link_libraries(commissioner
         fmt::fmt
         event_core
         event_pthreads
+        $<$<NOT:$<BOOL:${OT_COMM_ANDROID}>>:pthread>
```
wgtdkp, line 89: Android build pthread into its libc, so there is no separate `pthread` to link.

Issue #28, opened 2020-03-11
No error provided in response to a bad PSKd


jwhui:
An error would be useful to inform the user.

wgtdkp:
Agree.

PR #95, opened 2020-05-04
[api] refactor commissioner handlers

wgtdkp:
This PR refactors commissioner handlers to use a `CommissionerHandler` interface for handling all commissioner events. An application should implement the `CommisssionerHandler` interface to handle those events.

Change summary:
- use `CommissionerHandler` interface to handle commissioner events.
- add `OnJoinerConnected` to handle joiner DTLS connection event. This addresses issue https://github.com/openthread/ot-commissioner/issues/28.
- add `OnKeepAliveResponse`  to notify whether the commissioner was rejected or not.
- remove `JoinerType` and `JoinerInfo` from public library interface.

This PR supports Java binding of the commissioner library.

### DIFF
--- a/include/commissioner/commissioner.hpp
+++ b/include/commissioner/commissioner.hpp
@@ -208,6 +208,18 @@ public:
     }
 
     /**
+     * This funtions notifies the response of a keep-alive message.
+     *
+     * @param[in] aError  An error indicates whether the keep-alive message
+     *                    was accepted by the leader.
+     *
+     */
+    virtual void OnKeepAliveResponse(Error aError)
+    {
+        (void)aError;
+    }
+
+    /**
      * This function notifies the receiving a PAN ID conflict answer.
      *
      * @param[in] aPeerAddr     A peer address that sent the MGMT_PANID_CONFLICT.ans request.
--- a/include/commissioner/commissioner.hpp
+++ b/include/commissioner/commissioner.hpp
@@ -214,10 +214,7 @@ public:
      *                    was accepted by the leader.
      *
      */
-    virtual void OnKeepAliveResponse(Error aError)
-    {
-        (void)aError;
-    }
+    virtual void OnKeepAliveResponse(Error aError) { (void)aError; }
 
     /**
      * This function notifies the receiving a PAN ID conflict answer.
--- a/include/commissioner/commissioner.hpp
+++ b/include/commissioner/commissioner.hpp
@@ -217,7 +217,7 @@ public:
      * @param[in] aPanId        The PAN ID that has a conflict.
      *
      */
-    virtual void OnPanIdConflict(const std::string &aPeerAddr, const ChannelMask &aChannelMask, const uint16_t &aPanId)
+    virtual void OnPanIdConflict(const std::string &aPeerAddr, const ChannelMask &aChannelMask, uint16_t aPanId)
     {
         (void)aPeerAddr;
         (void)aChannelMask;
@@ -263,7 +263,7 @@ public:
     /**
      * The response handler of a general TMF request.
      *
-     * @param[in] aError  A error code.
+     * @param[in] aError  An error code.
      */
     using ErrorHandler = std::function<void(Error aError)>;
 
@@ -271,12 +271,12 @@ public:
      * The response handler of a general TMF request.
      *
      * @param[in] aResponseData  A response data. nullable.
-     * @param[in] aError         A error code.
+     * @param[in] aError         An error code.
      *
      * @note @p aResponseData is guaranteed to be not null only when @p aError == Error::kNone.
      *       Otherwise, @p aResponseData should never be accessed.
      */
-    template <class T> using Handler = std::function<void(const T *aResponseData, Error aError)>;
+    template <typename T> using Handler = std::function<void(const T *aResponseData, Error aError)>;
 
     /**
      * The petition result handler.
@@ -293,7 +293,7 @@ public:
      * @brief Create an instance of the commissioner.
      *
      * @param[out] aCommissioner  The created commissioner instance.
-     * @param[in] aHandler    A handler of commissioner events.
+     * @param[in]  aHandler       A handler of commissioner events.
      * @param[in]  aConfig        A commissioner configuration.
      *
      * @retval Error::kNone  Successfully created a commissioner instance.
@@ -1000,12 +1000,12 @@ public:
      * by sending MLR.req message to the Primary Backbone Router.
      * It will not return until errors happened, timeouted or succeed.
      *
-     * @param[out] aStatus            The Status.
-     * @param[in] aPbbrAddr           A Primary Backbone Router Address that the MLR.req message
-     *                                will be sent to.
-     * @param[in] aMulticastAddrList  A list of multicast address to be registered.
-     * @param[in] aTimeout            A time period after which the registration as a
-     *                                listener to the included multicast group(s) expires; In seconds.
+     * @param[out] aStatus             The Status.
+     * @param[in]  aPbbrAddr           A Primary Backbone Router Address that the MLR.req message
+     *                                 will be sent to.
+     * @param[in]  aMulticastAddrList  A list of multicast address to be registered.
+     * @param[in]  aTimeout            A time period after which the registration as a
+     *                                 listener to the included multicast group(s) expires; In seconds.
      * @return Error::kNone, succeed, the address has been successfully registered; Otherwise, failed;
      */
     virtual Error RegisterMulticastListener(uint8_t &                       aStatus,
--- a/include/commissioner/commissioner.hpp
+++ b/include/commissioner/commissioner.hpp
@@ -46,8 +46,6 @@
 #include <commissioner/error.hpp>
 #include <commissioner/network_data.hpp>
 
-struct event_base;
-
 namespace ot {
 
 namespace commissioner {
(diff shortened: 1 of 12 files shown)


include/commissioner/commissioner.hpp
```diff
@@ -127,35 +125,137 @@ struct Config
 };
 
 /**
- * @brief Enumeration of Joiner Type for steering.
+ * @brief The base class defines Handlers of commissioner events.
+ *
+ * Application should inherit this class and override the virtual
+ * functions to provide specific handler.
+ *
+ * @note keep in mind that those handlers will be called in another threads.
```
simonlingoogle, line 133:
```suggestion
 * @note keep in mind that those handlers will be called in the commissioner thread.
```

Better to explicitly call out the executing thread of handlers?

include/commissioner/commissioner.hpp
```diff
@@ -127,35 +125,137 @@ struct Config
...
+    /**
+     * The function notifies the start of a joining request from given joiner.
+     *
+     * @param[in]  aJoinerId  A joiner ID.
+     *
+     * @return PSKd of the joiner. An empty PSKd indicates that the joiner is not
+     *         enabled.
+     *
+     */
+    virtual std::string OnJoinerRequest(const ByteArray &aJoinerId)
```
simonlingoogle, line 148:
```suggestion
    virtual std::string OnJoinerRequest(const ByteArray &aJoinerId) = 0;
```

What about use abstract functions so that users have to override these methods?

include/commissioner/commissioner.hpp
```diff
@@ -127,35 +125,137 @@ struct Config
...
+     * @param[in]  aProvisioningUrl    A URL encoded as a utf-8 string provided by the Joiner
+     *                                 to communicate to the user which Commissioning application
+     *                                 is best suited to properly provision it to the appropriate
+     *                                 service. Empty if the joiner doesn't provide it.
+     * @param[in]  aVendorData         A product vendor-defined data structure to guide
+     *                                 vendor-specific provisioning. Empty if the joiner doesn't provide it.
+     *
+     * @return  A boolean indicates whether the joiner is accepted.
+     *
+     * @note This will be called when A well-formed JOIN_FIN.req has been received.
```
simonlingoogle, line 188:
```suggestion
     * @note This will be called when a well-formed JOIN_FIN.req has been received.
```

include/commissioner/commissioner.hpp
```diff
@@ -127,35 +125,137 @@ struct Config
...
+        (void)aVendorData;
+        return false;
+    }
 
-    // If the value is all-zeros, it represents for all joiners of this type.
-    uint64_t mEui64; ///< The IEEE EUI-64 value.
+    /**
+     * This funtions notifies the response of a keep-alive message.
+     *
+     * @param[in] aError  An error indicates whether the keep-alive message
```
simonlingoogle, line 211: Clearly states what errors to expect?

include/commissioner/commissioner.hpp
```diff
@@ -194,90 +294,16 @@ class Commissioner
...
      * @return std::shared_ptr<Commissioner>  nullptr is returned if failed.
      *
-     * @note If @p aEventBase is not provided, commissioner will create it by itself and
-     *       run the event loop in separate background thread. In this case, all callback
-     *       function will be called in separate thread and synchronization is needed;
-     *       Otherwise, it is the user to run the event loop and no synchronization is
-     *       needed in callback functions.
      */
-    static std::shared_ptr<Commissioner> Create(const Config &aConfig, struct event_base *aEventBase);
+    static std::shared_ptr<Commissioner> Create(CommissionerHandler &aHandler, const Config &aConfig);
```
simonlingoogle, line 306:
```suggestion
    static std::shared_ptr<Commissioner> Create(const CommissionerHandler &aHandler, const Config &aConfig);
```

include/commissioner/commissioner.hpp
```diff
@@ -194,90 +294,16 @@ class Commissioner
      */
     using PetitionHandler = std::function<void(const std::string *aExistingCommissionerId, Error aError)>;
```
simonlingoogle, line 295: Why is `PetitionHandler` not also handled by `CommissionerHandler` ?

src/app/commissioner_app.cpp
```diff
@@ -1220,47 +1196,92 @@ void CommissionerApp::MergeDataset(CommissionerDataset &aDst, const Commissioner
...
+    // Dummy handler.
+}
+
+void CommissionerApp::OnPanIdConflict(const std::string &aPeerAddr, const ChannelMask &aChannelMask, uint16_t aPanId)
+{
+    (void)aPeerAddr;
     // Main thread will wait for updates to mPanIdConflicts,
     // which guarantees no concurrent access to it.
-    mPanIdConflicts[*aPanId] = *aChannelMask;
+    mPanIdConflicts[aPanId] = aChannelMask;
```
simonlingoogle, line 1268: How do we guarantee there is no race condition? I can see that `mPanIdConflicts` is accessed in CLI.

https://github.com/openthread/ot-commissioner/blob/d6ec82410e4da1263057a15ecc20bc74ea0e3f7a/src/app/cli/interpreter.cpp#L978

src/app/commissioner_app.cpp
```diff
@@ -1220,47 +1196,92 @@ void CommissionerApp::MergeDataset(CommissionerDataset &aDst, const Commissioner
...
-    SuccessOrExit(addr.Set(*aPeerAddr));
+    ASSERT(addr.Set(aPeerAddr) == Error::kNone);
 
     // Main thread will wait for updates to mPanIdConflicts,
     // which guarantees no concurrent access to it.
-    mEnergyReports[addr] = {*aChannelMask, *aEnergyList};
-
-exit:
-    return;
+    mEnergyReports[addr] = {aChannelMask, aEnergyList};
```
simonlingoogle, line 1280: Again, how to guarantee ?

src/app/commissioner_app.hpp
```diff
@@ -62,7 +61,37 @@ struct EnergyReport
 };
 using EnergyReportMap = std::map<Address, EnergyReport>;
 
-class CommissionerApp
+/**
+ * @brief Enumeration of Joiner Type for steering.
```
simonlingoogle, line 65:
```suggestion
 * @brief Enumeration of Joiner types for steering.
```

include/commissioner/commissioner.hpp
wgtdkp: The user may ask what is the **commissioner thread**. I think we'd better hide how we implement the commissioner interface. But, a user must know that the callback is not called in the user's main thread and he/she must add synchronization in the callback if any user data is accessed at there. In this case, whether the **another thread** is **commissioner thread** doesn't matter to the user but will add one more question to him/her. Thoughts?

include/commissioner/commissioner.hpp
wgtdkp: There are too concerns for not using pure virtual functions:
1. The user is very likely to define this dummy handler if he/she is not going to use Thread 1.1 Commissioning (for  example, the user just want the Thread Network Management functionalities).
2. When binding to java, technically, we cannot enforce a Java class implementing every virtual functions at here. There will be a segment fault If we define them pure virtual here but a Java class doesn't implement them.

include/commissioner/commissioner.hpp
wgtdkp: Giving the accurate errors is not easy. Please take a look at the call site of this handler https://github.com/openthread/ot-commissioner/pull/95/files#diff-664c9f59b79e0e70d3f1e8d022c9b49bR1151.

It propagates every error on the call chain, and we need to dig into every function called in the chain to aggregate a set of possible errors and this set will change when implementation changes.

Most of the errors in the commissioner interface serve indicating whether an operation has succeed or failed. This may not be the beast practice and we can discuss about it.

For this callback, we can normalize all possible errors to `Error::kNone` and `Error::kReject`, but it will drop the actual error, which can carry valuable debug information, during the keep-alive request-response exchange. Thoughts?

include/commissioner/commissioner.hpp
wgtdkp: Because petition is initiated by the user with the petition API:
```c++
virtual void Petition(PetitionHandler aHandler, const std::string &aAddr, uint16_t aPort) = 0;
```

Is it better to specify the callback handler when the user invoking the trigger?

src/app/commissioner_app.cpp
wgtdkp: The comment is stale, we have no protection against those data with current CLI implementation.

Will update the comment and add a lock for those data.

include/commissioner/commissioner.hpp
simonlingoogle: For 1), it does not seems to be a serious issue to me. Nowadays, most IDEs can generate default implementations for unimplemented methods. This also have the advantage that the user can explicitly see what callbacks do `CommissionerHandler` provide.

For 2), can we just make sure all pure virtual functions are implemented since we own the Java code?

include/commissioner/commissioner.hpp
simonlingoogle: Fine to not list all possible errors.

include/commissioner/commissioner.hpp
simonlingoogle: Seems reasonable to me.

include/commissioner/commissioner.hpp
simonlingoogle: One concern is that users can **NOT** tell from this that they should keep the handlers short (exit quickly) since `another thread` does not necessarily block the commissioner events.

include/commissioner/commissioner.hpp
wgtdkp: We cannot make it const, because those handlers are not `const` methods.

include/commissioner/commissioner.hpp
wgtdkp: > can we just make sure all pure virtual functions are implemented since we own the Java code?

We don't provide a `CommissionerHandler` implementation in Java. But I turns to agree we should make them pure virtual.

include/commissioner/commissioner.hpp
wgtdkp: Well, I can document **what should not be done in those callbacks**.

wgtdkp: @simonlingoogle the docstring is updated, please review.

src/app/commissioner_app.cpp
simonlingoogle: Not in this PR?

wgtdkp: since this is not directly related to this PR, let's do it in a separate PR.

src/app/commissioner_app.cpp
wgtdkp: fix in separate PR.

PR #98, opened 2020-05-04
[api] remove API 'start' and 'stop'

wgtdkp:
This PR removes API `start` and `stop`. Those two APIs are specific to event loop thread, which should be hide behind the public interface.

### DIFF
--- a/src/library/commissioner_safe.cpp
+++ b/src/library/commissioner_safe.cpp
@@ -35,6 +35,8 @@
 
 #include <future>
 
+#include <assert.h>
+
 #include "library/coap.hpp"
 #include "library/cose.hpp"
 #include "library/logging.hpp"
@@ -95,7 +97,7 @@ const Config &CommissionerSafe::GetConfig() const
 
 void CommissionerSafe::StartEventLoopThread()
 {
-    VerifyOrDie(!mEventThread.joinable());
+    assert(!mEventThread.joinable());
 
     mEventThread = std::thread([this]() {
         LOG_INFO("event loop started in background thread");
--- a/src/library/commissioner_safe.cpp
+++ b/src/library/commissioner_safe.cpp
@@ -100,7 +100,7 @@ void CommissionerSafe::StartEventLoopThread()
     assert(!mEventThread.joinable());
 
     mEventThread = std::thread([this]() {
-        LOG_INFO("event loop started in background thread");
+        LOG_INFO(LOG_REGION_MESHCOP, "event loop started in background thread");
         event_base_loop(mEventBase.Get(), EVLOOP_NO_EXIT_ON_EMPTY);
     });
 }
--- a/src/library/commissioner_safe.cpp
+++ b/src/library/commissioner_safe.cpp
@@ -76,13 +76,16 @@ Error CommissionerSafe::Init(const Config &aConfig)
     VerifyOrExit(event_assign(&mInvokeEvent, mEventBase.Get(), -1, EV_PERSIST, Invoke, this) == 0);
     VerifyOrExit(event_add(&mInvokeEvent, nullptr) == 0);
 
+    // Do not propogate the returned error.
+    SuccessOrExit(StartEventLoopThread());
+
 exit:
     return error;
 }
 
 CommissionerSafe::~CommissionerSafe()
 {
-    Stop();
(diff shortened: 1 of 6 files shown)


src/library/commissioner_safe.cpp
```diff
@@ -76,13 +76,16 @@ Error CommissionerSafe::Init(const Config &aConfig)
     VerifyOrExit(event_assign(&mInvokeEvent, mEventBase.Get(), -1, EV_PERSIST, Invoke, this) == 0);
     VerifyOrExit(event_add(&mInvokeEvent, nullptr) == 0);
 
+    // Do not propogate the returned error.
+    SuccessOrExit(StartEventLoopThread());
```
simonlingoogle, line 80: Could creating thread ever fail ? If it failed, can we just return Success?

src/library/commissioner_safe.cpp
```diff
@@ -91,20 +94,22 @@ const Config &CommissionerSafe::GetConfig() const
     return mImpl.GetConfig();
 }
 
-Error CommissionerSafe::Start()
+Error CommissionerSafe::StartEventLoopThread()
 {
     Error error = Error::kNone;
 
     VerifyOrExit(!mEventThread.joinable(), error = Error::kAlready);
```
simonlingoogle, line 101:
```suggestion
    ASSERT(!mEventThread.joinable());
```
Can we use ASSERT now since it can only be called once ?

src/library/commissioner_safe.cpp
wgtdkp: @simonlingoogle 
`StartEventLoopThread` will fail only if There has already been an "event loop Thread". We should avoid ignoring error like this. What about making `StartEventLoopThread` returns `void` and use `VerifyOrDie` inside it to make sure the developer is not using it wrong ( `StartEventLoopThread` should be only called when a new commissioner is created).

src/library/commissioner_safe.cpp
wgtdkp: I have the same idea in previous comment. Will do it.

wgtdkp: @simonlingoogle 
changed `StartEventLoopThread` to return `void`, please review.

src/library/commissioner_safe.cpp
simonlingoogle: > `StartEventLoopThread` will fail only if There has already been an "event loop Thread". 

Not very familiar with C++ `std::thread`, however I think it is always possible to unable to create a new thread from OS's perspective (maybe due to resource limits). I think in these situations, thread constructor will just fail with exceptions. Do you think it's necessary to convert this exception to `Error`?

wgtdkp: Yes, you are. The overall strategy of handling exceptions from STL is not handling it and let it die when an exception is thrown. Just like we don't catch exceptions on `std::make_shared<...>(...)`.

src/library/commissioner_safe.cpp
```diff
@@ -91,20 +93,17 @@ const Config &CommissionerSafe::GetConfig() const
...
 
-Error CommissionerSafe::Start()
+void CommissionerSafe::StartEventLoopThread()
 {
-    Error error = Error::kNone;
-
-    VerifyOrExit(!mEventThread.joinable(), error = Error::kAlready);
-
-    mEventThread = std::thread([this]() { IgnoreError(mImpl.Start()); });
+    VerifyOrDie(!mEventThread.joinable());
```
simonlingoogle, line 98: We are always using `VerifyOrDie` instead of `Assert`. However, I think this is one of the situations that we should use `Assert` because it is a logical error, rather than some unrecoverable system error. Thought?

wgtdkp: agree.

src/library/commissioner_safe.hpp
```diff
@@ -217,6 +214,9 @@ class CommissionerSafe : public Commissioner
         struct event_base *mEventBase;
     };
 
+    // The EventBaseHolder needs to be the first member so that
```
simonlingoogle, line 217: any reason why so ?

wgtdkp: Because any other modules depends on an the `event_base` object.

PR #102, opened 2020-05-04
[logging] add region tag to log messages

wgtdkp:
This PR
1. adds log region.
2. rename `CommissioningSession` to `JoinerSession`.

### DIFF
--- a/src/library/coap.cpp
+++ b/src/library/coap.cpp
@@ -497,7 +497,8 @@ void Coap::HandleRequest(const Request &aRequest)
     response = mResponsesCache.Match(aRequest);
     if (response != nullptr)
     {
-        LOG_INFO("found cached CoAP response for resource {}", uriPath);
+        LOG_INFO(LOG_REGION_COAP, "server(={}) found cached CoAP response for resource {}", static_cast<void *>(this),
+                 uriPath);
         ExitNow(error = Send(*response));
     }
 
@@ -521,7 +522,8 @@ void Coap::HandleRequest(const Request &aRequest)
 exit:
     if (error != Error::kNone)
     {
-        LOG_INFO("CoAP: handle request failed: %s", ErrorToString(error));
+        LOG_INFO(LOG_REGION_COAP, "server(={}) handle request failed: %s", static_cast<void *>(this),
+                 ErrorToString(error));
     }
     return;
 }
@@ -596,7 +598,7 @@ void Coap::Retransmit(Timer &)
 {
     auto now = Clock::now();
 
-    LOG_DEBUG("CoAP retransmit timer triggered");
+    LOG_DEBUG(LOG_REGION_COAP, "client(={}) retransmit timer triggered", static_cast<void *>(this));
 
     while (!mRequestsCache.IsEmpty() && mRequestsCache.Earliest() < now)
     {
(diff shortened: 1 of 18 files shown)


src/library/coap.cpp
```diff
@@ -521,7 +522,8 @@ void Coap::HandleRequest(const Request &aRequest)
 exit:
     if (error != Error::kNone)
     {
-        LOG_INFO("CoAP: handle request failed: %s", ErrorToString(error));
+        LOG_INFO(LOG_REGION_COAP, "server(={}) handle request failed: %s", static_cast<void *>(this),
```
simonlingoogle, line 525: why is there a mixed style of `{}` and `%s`

wgtdkp: `%s` is the stale format, will fix.

src/library/commissioner_impl.cpp
```diff
@@ -1156,7 +1157,7 @@ void CommissionerImpl::SendPetition(PetitionHandler aHandler)
 
     mBrClient.SendRequest(request, onResponse);
 
-    LOG_DEBUG("sent petition request");
+    LOG_DEBUG(LOG_REGION_GENERIC, "sent petition request");
```
jwhui, line 1160: Prefer to avoid the "GENERIC" name - it is too easy for this to be a catch all for any new logs. Can we use a more specific name here? How is this different than MESHCOP?

src/library/commissioning_session.cpp
```diff
@@ -114,8 +114,10 @@ Error CommissioningSession::SendRlyTx(const ByteArray &aDtlsMessage, bool aInclu
 
     mCommImpl.mBrClient.SendRequest(rlyTx, nullptr);
 
-    LOG_DEBUG("sent RLY_TX.ntf: CommissioningSessionState={}, joinerIID={}, length={}, includeKek={}",
-              mDtlsSession->GetStateString(), utils::Hex(GetJoinerIid()), aDtlsMessage.size(), aIncludeKek);
+    LOG_DEBUG(LOG_REGION_MESHCOP,
```
jwhui, line 117: MESHCOP seems a superset of MGMT. How about naming this region something like SESSION?

src/library/commissioner_impl.cpp
wgtdkp: oh, it seems I misunderstood the scope of MeshCoP and we should use `LOG_REGION_MESHCOP` instead.

src/library/commissioning_session.cpp
wgtdkp: It may be confusing to name it only `SESSION`, because there is a `Commissioner Session` connected with the BA. How about `LOG_REGION_COMMISSIONING_SESSION`? 

en... it is easy to be confused with `Commissioning Session` and `Commissioner Session`, I am thinking about renaming `commissioning_session` to `joiner_session`. Thoughts? @jwhui @simonlingoogle

jwhui: The Thread Specification differentiates as:

1. Commissioner Session: secure session between Commissioner and Border Agent.

2. Joiner Session: secure session between Commissioner and Joiner.

wgtdkp: Thanks, so I think we should rename `commissioning_session` (the session of commissioning a joiner) here to `joiner_session`. Agree?

jwhui: Sounds good to me!

wgtdkp: done.

src/library/commissioner_impl.cpp
wgtdkp: done.

PR #104, opened 2020-05-09
[continuous-integration] remove Travis CI

wgtdkp:
This PR removes Travis CI since we have GitHub Action now.


wgtdkp:
There is significant decrease of code coverage, investigating.

wgtdkp:
> There is significant decrease of code coverage, investigating.

**It seems the codecov bash script is counting coverage incorrectly**. we can take this [timer.hpp](https://codecov.io/gh/openthread/ot-commissioner/src/c2c0367c29f0e529369d0dd7004cdba9649b0bde/src/library/timer.hpp) report for example, the destructor is executed while the constructor is never executed. This is not possible.

The new commit https://github.com/openthread/ot-commissioner/pull/104/commits/26eeeac374729d852e9819f9b27a95bae53f962f switch to generate coverage summary at local and feed it to the codecov bash script. The [timer.hpp](https://codecov.io/gh/openthread/ot-commissioner/src/26eeeac374729d852e9819f9b27a95bae53f962f/src/library/timer.hpp) report now looks fine.

Not sure why the codecov bash script works well on travis but not on GitHub Action.

PR #105, opened 2020-05-11
[code] remove throws

wgtdkp:
This PR removes all `throws` but keep only `SuccessOrThrow` in `json.cpp`, since the json library uses `exception` exclusively.

By searching `throw` in `src`, it gives only one match in [json.cpp](https://github.com/openthread/ot-commissioner/blob/560e6025d30390999ead96f1f3deda78bf5e5ee1/src/app/json.cpp#L83) (called by `SuccessOrThrow`).

https://github.com/openthread/ot-commissioner/blob/560e6025d30390999ead96f1f3deda78bf5e5ee1/src/app/json.cpp#L78-L85

### DIFF
--- a/src/library/commissioner_safe.cpp
+++ b/src/library/commissioner_safe.cpp
@@ -60,9 +60,9 @@ exit:
     return error;
 }
 
-Error CommissionerSafe::Init(CommissionerHandler & aHandler, const Config &aConfig)
+Error CommissionerSafe::Init(CommissionerHandler &aHandler, const Config &aConfig)
 {
-    Error error = Error::kFailed;
+    Error                             error = Error::kFailed;
     std::shared_ptr<CommissionerImpl> impl;
 
     VerifyOrExit(mEventBase.Get() != nullptr, error = Error::kOutOfMemory);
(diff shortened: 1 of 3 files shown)


src/library/commissioner_safe.cpp
```diff
@@ -562,10 +562,7 @@ CommissionerSafe::AsyncRequest CommissionerSafe::PopAsyncRequest()
 CommissionerSafe::EventBaseHolder::EventBaseHolder()
     : mEventBase(event_base_new())
 {
-    if (mEventBase == nullptr)
```
simonlingoogle, line 565: exception not allowed even when memory error ?

wgtdkp: I think so, we should avoid `throw` when possible. Here we want to just terminate the program, and `VerifyOrDie` can do this for us.

simonlingoogle: I think is not proper to just **EXIT** the process when creating `CommissionerSafe` failed. 
The user of `CommissionerSafe` would expect following code to return nil when failed ?

```c++
  CommissionerSafe *c = new CommissionerSafe();
```

wgtdkp: Changed to check the failure in `Init`, please review @simonlingoogle

Issue #42, opened 2020-03-19
[ccm] The SetToken API will clear inner COM_TOK when failed.


wgtdkp:
current implementation of `SetToken` will clear `TokenManager`'s member data when setting a new Commissioner Token failed:

https://github.com/openthread/ot-commissioner/blob/037185556ec2d5ef82624af6d38a9433b7134ba0/src/library/token_manager.cpp#L242-L246

This is not an expected behavior for API user. We should keep the old Commissioner Token if failed to set a new one.

PR #107, opened 2020-06-02
[ccm] fix setting Commissioner Token

wgtdkp:
This PR addresses issue https://github.com/openthread/ot-commissioner/issues/42 that the Commissioner Token will be cleared when failed to set it by re-extracting the Token into a CBOR object after accepting the signed raw token.

### DIFF
--- a/src/library/token_manager.cpp
+++ b/src/library/token_manager.cpp
@@ -222,29 +222,35 @@ exit:
 Error TokenManager::SetToken(const ByteArray &aSignedToken, const mbedtls_pk_context &aPublicKey)
 {
     Error          error = Error::kNone;
-    CborMap        token;
+    ByteArray      oldSignedToken;
+    CborMap        oldToken;
     CborMap        cnf;
     CborMap        coseKey;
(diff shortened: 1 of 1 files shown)


src/library/token_manager.cpp
```diff
@@ -220,32 +221,30 @@ Error TokenManager::SetToken(const ByteArray &aSignedToken, const ByteArray &aCe
...
+    SuccessOrExit(error = VerifyToken(token, aSignedToken, aPublicKey));
+    SuccessOrExit(error = token.Get(cwt::kCnf, cnf));
     SuccessOrExit(error = cnf.Get(cwt::kCoseKey, coseKey));
     SuccessOrExit(error = coseKey.Get(cose::kKeyId, kid, kidLength));
 
-    error = Error::kNone;
+    mSignedToken = aSignedToken;
+
+    // We need to re-extract the token into a CBOR object, because a CBOR object
+    // directly reference to raw data in the signed-token buffer.
```
simonlingoogle, line 239:
```suggestion
    // directly references to raw data in the signed-token buffer.
```

PR #110, opened 2020-06-05
[socket] remove Socket::Reset() API

wgtdkp:
## Problem
The `DtlsSession` was trying to reset the `Socket` when disconnected but it shouldn't. Because the socket is not connected/bound by `DtlsSession`, so it should not be reset by `DtlsSession` as well.

## Solution
- removes resetting socket in `DtlsSession::Disconnect()` and remove `Socket::Reset()`.
- Frees socket fd when before connecting and binding to allow multiple calls to `Connect` and `Bind`.

### DIFF
--- a/src/library/socket.cpp
+++ b/src/library/socket.cpp
@@ -67,11 +67,6 @@ Socket::Socket(struct event_base *aEventBase)
 
 Socket::~Socket()
 {
-    Reset();
-}
-
-void Socket::Reset()
-{
     if (mEvent.ev_base != nullptr)
     {
         event_del(&mEvent);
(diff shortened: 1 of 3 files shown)


src/library/socket.cpp
```diff
@@ -120,6 +115,11 @@ int UdpSocket::Connect(const std::string &aHost, uint16_t aPort)
 {
     auto portStr = std::to_string(aPort);
 
+    if (mNetCtx.fd >= 0)
+    {
+        mbedtls_net_free(&mNetCtx);
+    }
```
simonlingoogle, line 121: nit: If `mbedtls_net_free` is idempotent, maybe we can just:
```suggestion
    mbedtls_net_free(&mNetCtx);
```

PR #112, opened 2020-06-09
[ccm] add compile-time option to exclude CCM features

wgtdkp:
This PR adds a CMake option `OT_COMM_CCM` and feature switch `OT_COMM_CONFIG_CCM_ENABLE` to turn off Commercial Commissioning Mode support. It has the benefits of increasing code coverage and remove unnecessary dependencies (cn-cbor, COSE-C) when `CCM` features is not needed.

### DIFF
--- a/src/library/CMakeLists.txt
+++ b/src/library/CMakeLists.txt
@@ -77,8 +77,8 @@ target_sources(commissioner
 
 target_link_libraries(commissioner
     PRIVATE
-        cn-cbor::cn-cbor
-        cose
+        $<$<BOOL:${OT_COMM_CCM}>:cn-cbor::cn-cbor>
+        $<$<BOOL:${OT_COMM_CCM}>:cose>
         mdns
         mbedtls
         mbedx509
@@ -117,8 +117,8 @@ install(TARGETS commissioner
 ## disabled for them (intended).
 if (BUILD_SHARED_LIBS)
     install(FILES
-            $<TARGET_FILE:cn-cbor::cn-cbor>
-            $<TARGET_FILE:cose>
+            $<$<BOOL:${OT_COMM_CCM}>:$<TARGET_FILE:cn-cbor::cn-cbor>>
+            $<$<BOOL:${OT_COMM_CCM}>:$<TARGET_FILE:cose>>
             $<TARGET_FILE:mdns>
             $<TARGET_FILE:mbedtls>
             $<TARGET_FILE:mbedx509>
@@ -165,8 +165,8 @@ if (OT_COMM_TEST)
     target_link_libraries(commissioner-test
         PRIVATE
             Catch2::Catch2
-            cn-cbor::cn-cbor
-            cose
+            $<$<BOOL:${OT_COMM_CCM}>:cn-cbor::cn-cbor>
+            $<$<BOOL:${OT_COMM_CCM}>:cose>
             mdns
             mbedtls
             mbedx509
--- a/src/library/CMakeLists.txt
+++ b/src/library/CMakeLists.txt
@@ -93,7 +93,7 @@ target_link_libraries(commissioner
 
 target_compile_definitions(commissioner
     PRIVATE
-        $<IF:$<BOOL:${OT_COMM_CCM}>, OT_COMM_CCM_ENABLE=1, OT_COMM_CCM_ENABLE=0>
+        $<IF:$<BOOL:${OT_COMM_CCM}>, OT_COMM_CONFIG_CCM_ENABLE=1, OT_COMM_CONFIG_CCM_ENABLE=0>
 )
 
 target_include_directories(commissioner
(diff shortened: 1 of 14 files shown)


src/library/CMakeLists.txt
```diff
@@ -90,6 +90,11 @@ target_link_libraries(commissioner
         commissioner-common
 )
 
+target_compile_definitions(commissioner
+    PRIVATE
+        $<IF:$<BOOL:${OT_COMM_CCM}>, OT_COMM_CCM_ENABLE=1, OT_COMM_CCM_ENABLE=0>
```
simonlingoogle, line 95: why not `OT_COMM_CONFIG_CCM_ENABLE`, following config macro naming convention of `openthread` ?

src/library/commissioner_impl.cpp
```diff
@@ -1161,10 +1183,17 @@ Error CommissionerImpl::SignRequest(coap::Request &aRequest, tlv::Scope aScope)
 
     VerifyOrExit(IsCcmMode(), error = Error::kInvalidState);
 
+#if OT_COMM_CCM_ENABLE
     SuccessOrExit(error = mTokenManager.SignMessage(signature, aRequest));
 
     SuccessOrExit(error = AppendTlv(aRequest, {tlv::Type::kCommissionerToken, mTokenManager.GetToken(), aScope}));
     SuccessOrExit(error = AppendTlv(aRequest, {tlv::Type::kCommissionerSignature, signature, aScope}));
+#else
+    (void)aRequest;
```
simonlingoogle, line 1192: unreachable code? should have `goto exit` by `VerifyOrExit(IsCcmMode(), error = Error::kInvalidState);`

src/library/commissioner_impl.cpp
```diff
@@ -1013,17 +1020,32 @@ void CommissionerImpl::RequestToken(Handler<ByteArray> aHandler, const std::stri
...
-    if (!IsCcmMode())
-    {
-        return Error::kInvalidState;
-    }
-    return mTokenManager.SetToken(aSignedToken, aSignerCert);
+    Error error;
+
+    VerifyOrExit(IsCcmMode(), error = Error::kInvalidState);
+
+#if OT_COMM_CCM_ENABLE
```
simonlingoogle, line 1039: `OT_COMM_CCM_ENABLE` should always be 1 here?

src/library/CMakeLists.txt
wgtdkp: okay to me, will do.

src/library/commissioner_impl.cpp
wgtdkp: `IsCcmMode` stands for `the Commissioner is configured to run in CCM mode`, I think we should reject a CCM configuration when `OT_COMM_CCM` is set to `OFF`.

simonlingoogle: yes, we should reject `CCM` configuration so that when `OT_COMM_CCM=0`, `IsCcmMode()` always returns `FALSE`. 
Thoughts?

src/library/commissioner_impl.cpp
```diff
@@ -149,7 +149,9 @@ CommissionerImpl::CommissionerImpl(CommissionerHandler &aHandler, struct event_b
     , mResourceUdpRx(uri::kUdpRx, [this](const coap::Request &aRequest) { mProxyClient.HandleUdpRx(aRequest); })
     , mResourceRlyRx(uri::kRelayRx, [this](const coap::Request &aRequest) { HandleRlyRx(aRequest); })
     , mProxyClient(mEventBase, mBrClient)
+#if OT_COMM_CONFIG_CCM_ENABLE
     , mTokenManager(mEventBase)
+#endif // OT_COMM_CONFIG_CCM_ENABLE
```
jwhui, line 154: nit: can avoid these comments when block is less than 20 lines

src/library/commissioner_impl.cpp
```diff
@@ -205,6 +209,11 @@ Error CommissionerImpl::ValidateConfig(const Config &aConfig)
     if (aConfig.mEnableCcm)
     {
         tlv::Tlv domainNameTlv{tlv::Type::kDomainName, aConfig.mDomainName};
+
+#if !OT_COMM_CONFIG_CCM_ENABLE
+        ExitNow(error = ERROR_INVALID_ARGS("CCM teatures not supported"));
```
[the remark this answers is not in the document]
wgtdkp: Thanks!

There is guideline that static objects with non-trivial destructor should be avoid. I defined an macro `CCM_NOT_IMPLEMENTED` to avoid repeating the error message:
```C++
#define CCM_NOT_IMPLEMENTED "CCM features not implemented"
```

src/library/commissioner_impl.cpp
wgtdkp: removed those comments.

PR #115, opened 2020-06-10
[dtls] handle events when disconnected

wgtdkp:
This PR fixes DTLS events handling when the session is disconnected. Previous implementation has the assumption that the underlying socket will be closed when the session is disconnected. But with PR https://github.com/openthread/ot-commissioner/pull/110,  the socket now may be open in this case.

This PR ignores the incoming data when the DTLS session is disconnected.

### DIFF
--- a/src/library/dtls.cpp
+++ b/src/library/dtls.cpp
@@ -403,7 +403,7 @@ void DtlsSession::HandleEvent(short aFlags)
     default:
         // Ignore incoming data when the DTLS session is disconnected.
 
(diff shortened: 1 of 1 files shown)


wgtdkp:
> Any background on why the underlying socket should not be closed when DTLS session is disconnected?

The `Socket` class is now applying RAII to manage its `fd` which means the socket is closed only when the socket object is destructed. This has the problem that we cannot reconnect to a Border Agent unless we release current `CoapSecure` (and `Socket`) object and create a new one. We need to add `Socket::Reset` back if we allow closing a Socket before destruction. Thoughts? @simonlingoogle

PR #116, opened 2020-06-12
[error] add unittest for Error class

wgtdkp:
This PR adds unittest for the `Error` class.

### DIFF
--- a/src/common/error_test.cpp
+++ b/src/common/error_test.cpp
@@ -1,0 +1,83 @@
+/*
+ *    Copyright (c) 2020, The OpenThread Commissioner Authors.
+ *    All rights reserved.
+ *
+ *    Redistribution and use in source and binary forms, with or without
+ *    modification, are permitted provided that the following conditions are met:
+ *    1. Redistributions of source code must retain the above copyright
+ *       notice, this list of conditions and the following disclaimer.
+ *    2. Redistributions in binary form must reproduce the above copyright
+ *       notice, this list of conditions and the following disclaimer in the
+ *       documentation and/or other materials provided with the distribution.
+ *    3. Neither the name of the copyright holder nor the
+ *       names of its contributors may be used to endorse or promote products
+ *       derived from this software without specific prior written permission.
+ *
+ *    THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
+ *    AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+ *    IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ *    ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE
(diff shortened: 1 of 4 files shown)


src/common/error_test.cpp
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+TEST_CASE("error-to-string", "[error]")
+{
+    SECTION("default Error object should return 'NONE' as string representation")
+    {
+        Error error;
+
+        REQUIRE(error.ToString() == "NONE");
```
simonlingoogle, line 59: - Use `OK` just like openthread?

wgtdkp: > Test string formatting for unsuccessful erros

Only `NONE` is documented for the `Error::ToString` API and string representation of other error are not documented and may change in the future. That means the user can reply on the truth that `Error(kNone).ToString()` always returns `NONE`, but no guarantee is provided for the rest errors.

> Test that error can be compared with error code
This has been tested by `TEST_CASE("error-comparison", "[error]")` at below.

> Test the behavior of comparing unsuccessful errors
I am not sure if we need to do, I am assuming that if comparison of one kind of error works, then the rest will work. Do we need to test if `ErrorA1 != ErrorB1`? There will be `17 * 17` testcases for this, and I think it is not necessary.

wgtdkp:
> Test string formatting for unsuccessful erros

Only NONE is documented for the Error::ToString API and string representation of other error are not documented and may change in the future. That means the user can reply on the truth that Error(kNone).ToString() always returns NONE, but no guarantee is provided for the rest errors.

> Test that error can be compared with error code

This has been tested by TEST_CASE("error-comparison", "[error]") at below.

> Test the behavior of comparing unsuccessful errors

I am not sure if we need to do, I am assuming that if comparison of one kind of error works, then the rest will work. Do we need to test if ErrorA1 != ErrorB1? There will be 17 * 17 testcases for this, and I think it is not necessary.

src/common/error_test.cpp
wgtdkp: ok ot me.

simonlingoogle:
> Only NONE is documented for the Error::ToString API and string representation of other error are not documented and may change in the future. That means the user can reply on the truth that Error(kNone).ToString() always returns NONE, but no guarantee is provided for the rest errors.

Can we check that the `Error::ToString` returns `ErrorCodeToString(mCode) + ": " + mMessage)` ?
 
> > Test that error can be compared with error code
> 
> This has been tested by `TEST_CASE("error-comparison", "[error]")`.
> 
> > Test the behavior of comparing unsuccessful errors
> 
> I am not sure if we need to do, I am assuming that if comparison of one kind of error works, then the rest will work. Do we need to test if ErrorA1 != ErrorB1? There will be 17 * 17 testcases for this, and I think it is not necessary.

The tests only test for `OK`, but it may easily be the case that `OK` are implemented differently from `non-OK` errors. I think we should at lease test for one `non-OK` error for code coverage.

wgtdkp:
> Can we check that the Error::ToString returns ErrorCodeToString(mCode) + ": " + mMessage) ?

I don't see a reason to test this, because we don't provide the format as a stable interface.

> The tests only test for OK, but it may easily be the case that OK are implemented differently from non-OK errors. I think we should at least test for one non-OK error for code coverage.

Since current implementation doesn't treat `OK` and `non-OK` errors, the error comparators are actually already covered by the unittest here. For future change that may implement `OK` error differently, it is the changer to write the additional unittest for it.

PR #121, opened 2020-06-22
[socket] fix deleting socket event when reconnect or rebind

wgtdkp:
The `Connect` and `Bind` methods are reentrant, we should delete the associated event struct if it has already been added to the event loop.


src/library/socket.cpp
```diff
@@ -117,6 +117,10 @@ int UdpSocket::Connect(const std::string &aHost, uint16_t aPort)
 
     // Free the fd if already opened.
     mbedtls_net_free(&mNetCtx);
+    if (mEvent.ev_base != nullptr)
+    {
+        event_del(&mEvent);
+    }
```
simonlingoogle, line 124: wrap the cleanup logic in a separate method?

PR #128, opened 2020-07-18
[event] fix ading user triggered events

wgtdkp:
This PR has two small fixes
1. Add a timeout value for the async call event so that the event loop will not exit because of no events.
2. User triggered event struct (with fd equals -1) should not listen on READ, WRITE events.

### DIFF
--- a/src/library/commissioner_safe.cpp
+++ b/src/library/commissioner_safe.cpp
@@ -54,13 +54,19 @@ Error CommissionerSafe::Init(const Config &aConfig)
     Error                             error;
     std::shared_ptr<CommissionerImpl> impl;
 
+    // The default timeout value (1 day) for non-IO events (events with fd < 0).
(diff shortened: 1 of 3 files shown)


src/library/commissioner_safe.cpp
```diff
@@ -89,7 +95,7 @@ void CommissionerSafe::StartEventLoopThread()
 
     mEventThread = std::thread([this]() {
         LOG_INFO(LOG_REGION_MESHCOP, "event loop started in background thread");
-        event_base_loop(mEventBase.Get(), EVLOOP_NO_EXIT_ON_EMPTY);
```
simonlingoogle, line 92: EVLOOP_NO_EXIT_ON_EMPTY not working?

wgtdkp: It can work for now. But it is a relatively new feature that some platform with an older `libevent` don't have support of this. We may downgrade the `libevent` library version later to support those platforms.

PR #129, opened 2020-07-23
[tests] enhance mDNS discover tests

wgtdkp:
This PR enhances the mDNS discovery tests
1. A full list of TXT values defined by Thread 1.2 ( and 1.1)
2. Fix the State bitmap format

### DIFF
--- a/src/app/border_agent.cpp
+++ b/src/app/border_agent.cpp
@@ -241,11 +241,11 @@ static int HandleRecord(const struct sockaddr *from,
                     ExitNow(error = ERROR_BAD_FORMAT("value of TXT Key 'sb' is invalid: value={}", utils::Hex(bitmap)));
                 }
 
-                borderAgent.mState.mConnectionMode = (bitmap[0] >> 5);
-                borderAgent.mState.mThreadIfStatus = (bitmap[0] << 3) >> 6;
-                borderAgent.mState.mAvailability   = (bitmap[0] << 5) >> 6;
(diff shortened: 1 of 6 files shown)


src/app/border_agent.cpp
```diff
@@ -241,11 +241,11 @@ static int HandleRecord(const struct sockaddr *from,
                     ExitNow(error = ERROR_BAD_FORMAT("value of TXT Key 'sb' is invalid: value={}", utils::Hex(bitmap)));
                 }
 
-                borderAgent.mState.mConnectionMode = (bitmap[0] >> 5);
-                borderAgent.mState.mThreadIfStatus = (bitmap[0] << 3) >> 6;
-                borderAgent.mState.mAvailability   = (bitmap[0] << 5) >> 6;
-                borderAgent.mState.mBbrIsActive    = (bitmap[0] << 7) >> 7;
-                borderAgent.mState.mBbrIsPrimary   = (bitmap[1] >> 7);
+                borderAgent.mState.mConnectionMode = (bitmap[3] & 0x07);
```
simonlingoogle, line 244: can we use a struct to better read each bits?

wgtdkp: If you are talking something about `borderAgent.mState = *reinterpret_cast<State*>(bitmap)`, then I would say this is not safe. Because the order of each bitfields is not defined by the C/C++ language, but implementation defined.

simonlingoogle: Fine to keep it as is

PR #130, opened 2020-07-23
[tests] add more negative tests for Address

wgtdkp:


src/common/address_test.cpp
```diff
@@ -132,6 +132,22 @@ TEST_CASE("address-negative-tests", "[address]")
 {
     Address addr;
 
+    SECTION("invalid address ToString should fail")
+    {
+        REQUIRE(addr.Set(addr.ToString()) == ErrorCode::kInvalidArgs);
```
simonlingoogle, line 137: What `addr.ToString()` returns?

wgtdkp: `Address::ToString()` will return `INVALID_ADDR` for invalid `Address` objects. This is to ease our usage of `Address::ToString()`, because in most cases, we call `Address::ToString()` to log peer address.

PR #132, opened 2020-07-23
[tests] add more testcases for error string

wgtdkp:


src/common/error_test.cpp
```diff
@@ -58,6 +60,32 @@ TEST_CASE("error-to-string", "[error]")
 
         REQUIRE(error.ToString() == "OK");
     }
+
+    SECTION("non-OK errors return its string representation (not required by the class Error API)")
```
simonlingoogle, line 64:
```suggestion
    SECTION("non-OK errors return their string representations (not required by the class Error API)")
```

src/common/error_test.cpp
```diff
@@ -58,6 +60,32 @@ TEST_CASE("error-to-string", "[error]")
...
+    {
+// Check if the error string starts with the error name.
+#define CHECK_ERROR_STR(aErrorName) REQUIRE(ERROR_##aErrorName("").ToString().find(#aErrorName) == 0)
```
simonlingoogle, line 67:
```suggestion
#define CHECK_ERROR_STR_STARTS_WITH_ERROR_NAME(aErrorName) REQUIRE(ERROR_##aErrorName("").ToString().find(#aErrorName) == 0)
```

PR #138, opened 2020-07-26
[cli] fix signal handling

wgtdkp:
Synchronization primitives (std::mutex, std::future) in namespace `std` are used to synchronize calls to the commissioner API. Unfortunately, those synchronization primitives are not [signal-safe](https://man7.org/linux/man-pages/man7/signal-safety.7.html) and will result in deadlock.

This PR fixes this by using a dedicated thread to wait for the signal.

### DIFF
--- a/src/app/cli/main.cpp
+++ b/src/app/cli/main.cpp
@@ -31,6 +31,8 @@
  *   The file is the entrance of the commissioner CLI.
  */
 
+#include <thread>
+
 #include <signal.h>
 
 #include "app/cli/interpreter.hpp"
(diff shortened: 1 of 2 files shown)


src/app/cli/main.cpp
```diff
@@ -76,30 +78,41 @@ static void PrintVersion()
...
-static void HandleSignalInterrupt(int)
+static void HandleSignalInterrupt()
 {
-    gInterpreter.AbortCommand();
+    int signalNum = 0;
+
+    while (true)
+    {
+        sigwait(&gSignalSet, &signalNum);
+        gInterpreter.AbortCommand();
```
simonlingoogle, line 90: should we quit this thread after this?

wgtdkp: This Thread is designed to run in background to handle the interrupt signals. As the user can input `CRTL + C` in the CLI for multiple times, we need to keep this thread running to handle more interrupts (The name `AbortCommand` is misleading, it should be `CancelCommand` and I have fired a PR https://github.com/openthread/ot-commissioner/pull/135 to rename it).

simonlingoogle: nit: If this thread goes on, you need to make sure that `AbortCommand` is reentrant.

wgtdkp: Sure, it is.

jwhui:
@wgtdkp , please help resolve conflicts.

wgtdkp:
@jwhui done.

src/app/cli/main.cpp
```diff
@@ -76,30 +78,41 @@ static void PrintVersion()
...
     {
         PrintVersion();
-        return 0;
+        ExitNow();
     }
 
-    signal(SIGINT, HandleSignalInterrupt);
+    // Block signals in this thread and subsequently spawned threads.
+    sigemptyset(&gSignalSet);
+    sigaddset(&gSignalSet, SIGINT);
```
bukepo, line 112: is SIGINT the only one should take care of?

PR #142, opened 2020-08-24
[active-dataset] returns raw Active Operational Dataset TLVs

wgtdkp:
Parsing the Active Operational Dataset into `struct ActiveOperationalDataset` is for manipulating each field of the dataset and the commissioner app doesn't need to write a decoder by their own. But the commissioner app may store or transfer Active Operational Dataset as a byte array. It would be convenient to have the raw TLV format of the dataset so that the commissioner app doesn't need to write an encoder by their own.

This PR adds a new APIs (`GetRawActiveDataset(...)`) to return raw Active Operational Dataset.

### DIFF
--- a/include/commissioner/commissioner.hpp
+++ b/include/commissioner/commissioner.hpp
@@ -641,7 +641,7 @@ public:
      * It always returns immediately without waiting for the completion.
      *
      * @param[in, out] aHandler       A handler of the response and errors; Guaranteed to be called.
-     * @param[in]      aDatasetFlags  Active Operational Dataset flags indicate which TLVs are wanted.
+     * @param[in]      aDatasetFlags  Active Operational Dataset flags indicating which TLVs are wanted.
      */
     virtual void GetRawActiveDataset(Handler<ByteArray> aHandler, uint16_t aDatasetFlags) = 0;
 
--- a/include/commissioner/commissioner.hpp
+++ b/include/commissioner/commissioner.hpp
@@ -632,6 +632,36 @@ public:
     virtual Error GetActiveDataset(ActiveOperationalDataset &aDataset, uint16_t aDatasetFlags) = 0;
 
     /**
+     * @brief Asynchronously get the raw Active Operational Dataset as a binary blob (in format of Thread TLV).
+     *
+     * Get the uninterpreted Active Operational Dataset in MGMT_ACTIVE_GET.rsp.
+     *
+     * This method request Active Operational Dataset of the Thread network
+     * by sending MGMT_ACTIVE_GET.req message.
+     * It always returns immediately without waiting for the completion.
+     *
+     * @param[in, out] aHandler       A handler of the response and errors; Guaranteed to be called.
+     * @param[in]      aDatasetFlags  Active Operational Dataset flags indicate which TLVs are wanted.
+     */
+    virtual void GetRawActiveDataset(Handler<ByteArray> aHandler, uint16_t aDatasetFlags) = 0;
+
+    /**
+     * @brief Synchronously get the raw Active Operational Dataset as a binary blob (in format of Thread TLV).
+     *
+     * Get the uninterpreted Active Operational Dataset in MGMT_ACTIVE_GET.rsp.
+     *
+     * This method request Active Operational Dataset of the Thread network
+     * by sending MGMT_ACTIVE_GET.req message.
+     * It will not return until errors happened, timeouted or succeed.
+     *
+     * @param[out] aDataset       A Active Operational Dataset returned by the leader.
+     * @param[in]  aDatasetFlags  Active Operational Dataset flags indicate which TLVs are wanted.
+     *
+     * @return Error::kNone, succeed; otherwise, failed;
+     */
+    virtual Error GetRawActiveDataset(ByteArray &aRawDataset, uint16_t aDatasetFlags) = 0;
+
+    /**
      * @brief Asynchronously set the Active Operational Dataset.
      *
      * This method set Active Operational Dataset of the Thread network
--- a/include/commissioner/commissioner.hpp
+++ b/include/commissioner/commissioner.hpp
@@ -636,7 +636,7 @@ public:
      *
      * Get the uninterpreted Active Operational Dataset in MGMT_ACTIVE_GET.rsp.
      *
-     * This method request Active Operational Dataset of the Thread network
+     * This method requests Active Operational Dataset of the Thread network
      * by sending MGMT_ACTIVE_GET.req message.
      * It always returns immediately without waiting for the completion.
      *
--- a/include/commissioner/commissioner.hpp
+++ b/include/commissioner/commissioner.hpp
@@ -650,7 +650,7 @@ public:
      *
      * Get the uninterpreted Active Operational Dataset in MGMT_ACTIVE_GET.rsp.
      *
-     * This method request Active Operational Dataset of the Thread network
+     * This method requests Active Operational Dataset of the Thread network
      * by sending MGMT_ACTIVE_GET.req message.
      * It will not return until errors happened, timeouted or succeed.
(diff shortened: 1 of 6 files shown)


wgtdkp:
@simonlingoogle @jwhui I will close this PR since CHIP is not going to use the Thread TLV format directly. 

See handling of Active Operational Dataset in CHIP: https://github.com/project-chip/connectedhomeip/blob/9754e3d0231d56aeb5495a580d6059cc1d2f1872/examples/common/chip-app-server/Server.cpp#L75-L140

jwhui:
> @simonlingoogle @jwhui Thanks for your comments, I will close this PR since CHIP is not going to use the Thread TLV format directly.
> 
> See handling of Active Operational Dataset in CHIP: https://github.com/project-chip/connectedhomeip/blob/9754e3d0231d56aeb5495a580d6059cc1d2f1872/examples/common/chip-app-server/Server.cpp#L75-L140

I think CHIP must treat the Active Operational Dataset as a binary blob. The Active Operational Dataset include fields defined by the Thread Specification and newer versions of the Thread Specification may introduce new fields. CHIP should be able to support newer versions of Thread without requiring software update of CHIP itself.

bukepo:
> > @simonlingoogle @jwhui Thanks for your comments, I will close this PR since CHIP is not going to use the Thread TLV format directly.
> > See handling of Active Operational Dataset in CHIP: https://github.com/project-chip/connectedhomeip/blob/9754e3d0231d56aeb5495a580d6059cc1d2f1872/examples/common/chip-app-server/Server.cpp#L75-L140
> 
> I think CHIP must treat the Active Operational Dataset as a binary blob. The Active Operational Dataset include fields defined by the Thread Specification and newer versions of the Thread Specification may introduce new fields. CHIP should be able to support newer versions of Thread without requiring software update of CHIP itself.

Should CHIP be aware of some of the fields? For example, when forming a new Thread network, CHIP may want to set the network name and channel mask. Thoughts?

jwhui:
CHIP can be aware of the fields, but CHIP must not exclude fields simply because it does not understand them.

wgtdkp:
> I think CHIP must treat the Active Operational Dataset as a binary blob. The Active Operational Dataset include fields defined by the Thread Specification and newer versions of the Thread Specification may introduce new fields. CHIP should be able to support newer versions of Thread without requiring software update of CHIP itself.

This makes sense! Thanks! @jwhui

src/library/commissioner_impl.cpp
```diff
@@ -1510,7 +1510,8 @@ Error CommissionerImpl::DecodeActiveOperationalDataset(ActiveOperationalDataset
         dataset.mPresentFlags |= ActiveOperationalDataset::kSecurityPolicyBit;
     }
 
-    aDataset = dataset;
+    dataset.mRawTlvs = aResponse.GetPayload();
```
wgtdkp, line 1513: @simonlingoogle I think we should prefer getting the binary Active Operational Dataset from the response directly instead of decoding and encoding it again. Decoding the Active Operational Dataset will filter out any TLVs not known by the Commissioner, while returning Active Operational Dataset binary keeps it unchanged and has better compatibility. Thoughts?

simonlingoogle, line 1513: can we be sure that there is no other TLV in the response?

simonlingoogle: It make sense to me to output raw bytes without dissecting it. 
The problem is that there should be no such field as `mRawTlvs` in in the ActiveOperationalDataset struct. 
Moreover, the logic itself is not **decodeing** so it should not be in the method `DecodeActiveOperationalDataset`. 

I would suggest to use a new method (e.g. GetRawActiveOperationalDataset(aResponse)) to get the raw bytes. 

Thoughts?

wgtdkp: Works for me.

wgtdkp: By Thread spec,  there is no other TLVs in the response.

wgtdkp: done as suggested, please review @simonlingoogle .

include/commissioner/commissioner.hpp
```diff
@@ -631,6 +631,36 @@ class Commissioner
      */
     virtual Error GetActiveDataset(ActiveOperationalDataset &aDataset, uint16_t aDatasetFlags) = 0;
 
+    /**
+     * @brief Asynchronously get the raw Active Operational Dataset as a binary blob (in format of Thread TLV).
+     *
+     * Get the uninterpreted Active Operational Dataset in MGMT_ACTIVE_GET.rsp.
+     *
+     * This method request Active Operational Dataset of the Thread network
```
simonlingoogle, line 639:
```suggestion
     * This method requests Active Operational Dataset of the Thread network
```

include/commissioner/commissioner.hpp
```diff
@@ -631,6 +631,36 @@ class Commissioner
...
+     * by sending MGMT_ACTIVE_GET.req message.
+     * It always returns immediately without waiting for the completion.
+     *
+     * @param[in, out] aHandler       A handler of the response and errors; Guaranteed to be called.
+     * @param[in]      aDatasetFlags  Active Operational Dataset flags indicate which TLVs are wanted.
```
simonlingoogle, line 644:
```suggestion
     * @param[in]      aDatasetFlags  Active Operational Dataset flags indicating which TLVs are wanted.
```

include/commissioner/commissioner.hpp
```diff
@@ -631,6 +631,36 @@ class Commissioner
...
+     */
+    virtual void GetRawActiveDataset(Handler<ByteArray> aHandler, uint16_t aDatasetFlags) = 0;
+
+    /**
+     * @brief Synchronously get the raw Active Operational Dataset as a binary blob (in format of Thread TLV).
+     *
+     * Get the uninterpreted Active Operational Dataset in MGMT_ACTIVE_GET.rsp.
+     *
+     * This method request Active Operational Dataset of the Thread network
```
simonlingoogle, line 653:
```suggestion
     * This method requests Active Operational Dataset of the Thread network
```

src/library/commissioner_impl.cpp
```diff
@@ -445,24 +445,41 @@ void CommissionerImpl::SetCommissionerDataset(ErrorHandler aHandler, const Commi
...
         SuccessOrExit(error = aError);
         VerifyOrExit(aResponse->GetCode() == coap::Code::kChanged,
                      error = ERROR_BAD_FORMAT("expect CoAP::CHANGED for MGMT_ACTIVE_GET.rsp message"));
 
-        SuccessOrExit(error = DecodeActiveOperationalDataset(dataset, *aResponse));
-        VerifyOrExit(dataset.mPresentFlags & ActiveOperationalDataset::kActiveTimestampBit,
-                     error = ERROR_BAD_FORMAT("Active Timestamp is not included in MGMT_ACTIVE_GET.rsp"));
-
-        aHandler(&dataset, error);
+        aHandler(&aResponse->GetPayload(), error);
```
simonlingoogle, line 482: According to Thread Spec., a message can contains extra TLVs and implementations are supposed to ignore these TLVs if unable to handle. So, I would suggest to only get the bytes of the Active Dataset TLV if it's easy to do.

simonlingoogle (approved):
LGTM 👍 

with nits

src/library/commissioner_impl.cpp
wgtdkp: Yes, extra TLVs are ignored when decoding the binary blob to struct `ActiveOperationalDataset`, but I think we should return what we get here, whether the receiver of the binary blob is going to ignore or reinterpret the extra TLVs is not limited by the commissioner.

jwhui (approved):

PR #143, opened 2020-09-02
[thci] output commissioning log and add thci interface to get it

wgtdkp:
This Adds a new API `getCommissioningLogs` to return decrypted contents of `JOIN_FIN.req/rsp` to the THCI program.

See the same THCI API in openthread: https://github.com/openthread/openthread/blob/002cc8e912a50abda089998c55c8c7f731b5cbf6/tools/harness-thci/OpenThread_WpanCtl.py#L2123


Moandor-y (approved):
LGTM

src/library/joiner_session.cpp
```diff
@@ -207,8 +208,17 @@ Error JoinerSession::SendJoinFinResponse(const coap::Request &aJoinFinReq, bool
     joinFin.SetSubType(MessageSubType::kJoinFinResponse);
     SuccessOrExit(error = mCoap.SendResponse(aJoinFinReq, joinFin));
 
+    LOG_INFO(LOG_REGION_JOINER_SESSION, "session(={}) sent JOIN_FIN.rsp: accepted={}", static_cast<void *>(this),
```
simonlingoogle, line 211: Could it be a security pitfall to print these messages as logs?

tools/commissioner_thci/commissioner_impl.py
```diff
@@ -401,6 +404,34 @@ def getCOM_TOK(self):
         result = self._execute_and_check('token print')
         return self._hex_to_bytes(result[0])
 
+    def getCommissioningLogs(self):
+        processed_logs = []
+        thci_logs = self._command("grep \"\\[ thci \\]\" {}".format(
+            self.log_file))
```
simonlingoogle, line 410: how do you clean `log_file` if we are running tests one by one?

tools/commissioner_thci/commissioner_impl.py
```diff
@@ -401,6 +404,34 @@ def getCOM_TOK(self):
         result = self._execute_and_check('token print')
         return self._hex_to_bytes(result[0])
 
+    def getCommissioningLogs(self):
```
simonlingoogle, line 407: `getCommissioningLogs` needs to wait for the commissioning process to finish. How do you make sure of that?

src/library/joiner_session.cpp
wgtdkp: What about make this optional with a compile flag?

tools/commissioner_thci/commissioner_impl.py
wgtdkp: The log file is overridden for each run of the commissioner.

tools/commissioner_thci/commissioner_impl.py
wgtdkp: I think we have no way to control when the method is called. The caller must know when to call this method.

src/library/joiner_session.cpp
simonlingoogle: Agree that a compile flag is needed. Just like in OT, we only print these logs when REFERENCE_DEVICE is ON.

tools/commissioner_thci/commissioner_impl.py
simonlingoogle: Not sure if that's true of TestHarness. We don't know how `getCommissioningLogs` is used by TestHarness.

.github/workflows/tests.yml
```diff
@@ -47,6 +47,7 @@ jobs:
                 -DCMAKE_BUILD_TYPE=Release        \
                 -DCMAKE_INSTALL_PREFIX=/usr/local \
                 -DOT_COMM_COVERAGE=ON             \
+                -DOT_COMM_COMMISSIONING_LOG=ON    \
```
simonlingoogle, line 50:
```suggestion
                -DOT_COMM_REFERENCE_DEVICE=ON    \
```

OT used to use `CERT_LOG` for certification logs, but later changed to `REFERENCE_DEVICE`.
Should we be consistent with OT and OTBR here?

simonlingoogle (approved):
LGTM :+1:
with a nit

.github/workflows/tests.yml
wgtdkp: Agree, will make the change accordingly.

.github/workflows/tests.yml
```diff
@@ -47,6 +47,7 @@ jobs:
                 -DCMAKE_BUILD_TYPE=Release        \
                 -DCMAKE_INSTALL_PREFIX=/usr/local \
                 -DOT_COMM_COVERAGE=ON             \
+                -DOT_COMM_REFERENCE_DEVICE=ON    \
```
jwhui, line 50: nit: alignment

CMakeLists.txt
```diff
@@ -29,13 +29,14 @@
...
-option(OT_COMM_APP              "Build the CLI App" ON)
-option(OT_COMM_CCM              "Build with Commercial Commissioning Mode" ON)
-option(OT_COMM_COVERAGE         "Enable coverage reporting" OFF)
-option(OT_COMM_JAVA_BINDING     "Build Java binding" OFF)
-set(OT_COMM_JAVA_BINDING_OUTDIR "" CACHE STRING "Specify output directory of generated Java source files")
-option(OT_COMM_TEST             "Build tests" ON)
+option(OT_COMM_ANDROID           "Build with Android NDK" OFF)
+option(OT_COMM_APP               "Build the CLI App" ON)
+option(OT_COMM_CCM               "Build with Commercial Commissioning Mode" ON)
+option(OT_COMM_REFERENCE_DEVICE "Enable features for being a Thread Group reference device" OFF)
```
jwhui, line 35: nit: alignment

jwhui (approved):

PR #149, opened 2020-10-10
[library] fix commissioner state when sending keep-alive failed

wgtdkp:
This PR fixes commissioner state transition when we failed to send the Keep-Alive message:
1. disconnect from the Border Agent.
2. Set the commissioner state to `kDisabled` when disconnected.

### DIFF
--- a/src/library/commissioner_impl.cpp
+++ b/src/library/commissioner_impl.cpp
@@ -318,6 +318,7 @@ void CommissionerImpl::Connect(ErrorHandler aHandler, const std::string &aAddr, 
 void CommissionerImpl::Disconnect()
 {
     mBrClient.Disconnect(ERROR_CANCELLED("the CoAPs client was disconnected"));
+    mState = State::kDisabled;
 }
 
 uint16_t CommissionerImpl::GetSessionId() const
@@ -1315,6 +1316,7 @@ exit:
     if (error != ErrorCode::kNone)
     {
         LOG_WARN(LOG_REGION_MESHCOP, "sending keep alive message failed: {}", error.ToString());
+        Disconnect();
     }
 }
 


librasungirl (approved):

simonlingoogle (commented):
LGTM 👍 

May I ask when would it happen that the commissioner failed to send KA ?

simonlingoogle (approved):

wgtdkp:
> LGTM 👍
> 
> May I ask when would it happen that the commissioner failed to send KA ?

@simonlingoogle The case I saw is in CCM mode. The Keep-Alive message is required to be signed and it may fail when an invalid `Commissioner Token` is used (this is allowed for certification).
https://github.com/openthread/ot-commissioner/blob/5cfcb390c5c0a7896974fcb6fb906b0733109cc1/src/library/commissioner_impl.cpp#L1302-L1305

There are also other cases that may fail this method (e.g. appending to the CoAP message failed) although they are very unusual.

simonlingoogle:
> > LGTM 👍
> > May I ask when would it happen that the commissioner failed to send KA ?
> 
> @simonlingoogle The case I saw is in CCM mode. The Keep-Alive message is required to be signed and it may fail when an invalid `Commissioner Token` is used (this is allowed for certification).
> https://github.com/openthread/ot-commissioner/blob/5cfcb390c5c0a7896974fcb6fb906b0733109cc1/src/library/commissioner_impl.cpp#L1302-L1305
> 
> There are also other cases that may fail this method (e.g. appending to the CoAP message failed) although they are very unusual.

Thanks for explanation. It makes sense to me to resign if sending KA failed.

jwhui (approved):

PR #150, opened 2020-10-12
[coap] fix retransmission of confirmable CoAP requests

wgtdkp:
This PR has two fixes to the CoAP retransmission of confirmable requests:
1. Stop the retransmission timer when all requests are cleared.
2. Start the retransmission timer when there are queueing requests that are waiting for responses.

### DIFF
--- a/src/library/coap.cpp
+++ b/src/library/coap.cpp
@@ -412,8 +412,6 @@ void Coap::CancelRequests()
         mRequestsCache.Front().mRequest->GetUriPath(uri).IgnoreError();
         FinalizeTransaction(mRequestsCache.Front(), nullptr, ERROR_CANCELLED("request to {} was cancelled", uri));
     }
-
-    mRequestsCache.StopTimer();
 }
 
 Error Coap::AddResource(const Resource &aResource)
@@ -863,6 +861,11 @@ void Coap::RequestsCache::Eliminate(const RequestHolder &aRequestHolder)
             mContainer.erase(holder);
             break;
         }
+    }
+
+    if (mContainer.empty())
+    {
+        mRetransmissionTimer.Stop();
     }
 }
 
--- a/src/library/coap.cpp
+++ b/src/library/coap.cpp
@@ -821,17 +821,7 @@ void Coap::RequestsCache::Put(const RequestPtr aRequest, ResponseHandler aHandle
 void Coap::RequestsCache::Put(const RequestHolder &aRequestHolder)
 {
     mContainer.emplace(aRequestHolder);
-    if (mRetransmissionTimer.IsRunning())
-    {
-        if (Earliest() < mRetransmissionTimer.GetFireTime())
-        {
-            mRetransmissionTimer.Start(Earliest());
-        }
-    }
-    else
-    {
-        mRetransmissionTimer.Start(Earliest());
-    }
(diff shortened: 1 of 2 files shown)


src/library/coap.cpp
```diff
@@ -412,6 +412,8 @@ void Coap::CancelRequests()
         mRequestsCache.Front().mRequest->GetUriPath(uri).IgnoreError();
         FinalizeTransaction(mRequestsCache.Front(), nullptr, ERROR_CANCELLED("request to {} was cancelled", uri));
     }
+
+    mRequestsCache.StopTimer();
```
librasungirl, line 416: should the retransmission timer be stopped in FinalizeTransaction() when there is no ongoing request entry?

if the response arrives, should the retransmissiontimer be stopped if there is no ongoing request? better to stop retransmission timer in

wgtdkp: Good point. Agree that it is better to handle this inner in the `RequestsCache` class. Updated.

src/library/coap.hpp
```diff
@@ -689,12 +689,9 @@ class Coap
...
 
-        // Clear all requests and stop retransmission timer.
-        void Clear();
-
-        // Try restart the retransmit timer if it is not running and there is
-        // pending requests.
-        void TryRetartTimer();
+        // Try starting the retransmit timer if it is not running
+        // and there is pending requests.
+        void TryStartTimer();
```
simonlingoogle, line 694:
```suggestion
        void UpdateTimer();
```

nit: `UpdateTimer` or  `ResetTimer` might be a better name. It can also stop timer if it's no longer needed.

src/library/coap.cpp
```diff
@@ -860,16 +862,14 @@ void Coap::RequestsCache::Eliminate(const RequestHolder &aRequestHolder)
...
-
-void Coap::RequestsCache::Clear()
-{
-    mRetransmissionTimer.Stop();
 
-    mContainer.clear();
+    if (mContainer.empty())
+    {
+        mRetransmissionTimer.Stop();
+    }
```
simonlingoogle, line 869:
```suggestion
    UpdateTimer()
```

We can simply put all timer starting and stopping logic to a single function (maybe called `UpdateTimer`) and call it whenever necessary.

librasungirl:
@wgtdkp would you please check the 2 failings?

wgtdkp:
> @wgtdkp would you please check the 2 failings?

@librasungirl resolved.

librasungirl (approved):

src/library/coap.cpp
```diff
@@ -860,18 +845,24 @@ void Coap::RequestsCache::Eliminate(const RequestHolder &aRequestHolder)
...
+    else if (mRetransmissionTimer.IsRunning())
+    {
+        if (Earliest() < mRetransmissionTimer.GetFireTime())
+        {
+            mRetransmissionTimer.Start(Earliest());
+        }
+    }
+    else
```
simonlingoogle, line 865: nit:
```suggestion
    else if (!mRetransmissionTimer.IsRunning() || Earliest() < mRetransmissionTimer.GetFireTime()) {
```

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #151, opened 2020-10-12
[swig] map Java byte to uin8_t for only the ByteArray class

wgtdkp:
This PR helps SWIG to map Java `byte` to/from C++ `uint8_t` when it is used for `std::vector<uint8_t>`.

SWIG by default maps Java `short` to/from C++ `uint8_t` because Java `byte` is a signed integer that cannot hold all values of `uint8_t`. But when it comes to a byte array, it doesn't matter whether it is signed or unsigned and we'd like to map `uint8_t`  to/from Java `byte` because they have the same width.

**The CHIP PR https://github.com/project-chip/connectedhomeip/pull/2492 depends on this (see the [discussion](https://github.com/project-chip/connectedhomeip/pull/2492#discussion_r503060646))**.

### DIFF
--- a/src/java/commissioner.i
+++ b/src/java/commissioner.i
@@ -44,12 +44,37 @@
 %}
 
 %include <std_string.i>
-%include <stdint.i>
 %include <std_list.i>
 %include <std_shared_ptr.i>
 %include <std_vector.i>
 %include <stl.i>
 %include <typemaps.i>
+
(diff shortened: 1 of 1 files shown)


simonlingoogle (approved):
LGTM 👍

jwhui (approved):

PR #152, opened 2020-10-16
[scripts] misc fixes in bootstrap script

wgtdkp:
This PR includes misc updates/fixes for bootstrapping:
1. downgrade required SWIG version to 3.0 since this is the default version that package tools have for Linux;
2. install SWIG in the bootstrap script;
3. install `python3` setup tools instead of `python2` since it is deprecated;

PR #153, opened 2020-10-22
[swig] revert changes to the minimum SWIG version

wgtdkp:
This PR reverts our changes of downgrading minimum SWIG version to 3.0 in #152.

We did downgrade the minimum version for easy installation of SWIG on old Linux distributions. But the incoming Commissioner Android App relies on the Java methods generated by SWIG 4.0. Unfortunately, we didn't have a GH job to test the SWIG-generated Java code for now, but it will come with the Android App PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -163,10 +163,7 @@ jobs:
           ninja
 
   java-binding:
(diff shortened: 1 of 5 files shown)


simonlingoogle (approved):
LGTM :+1:

PR #161, opened 2020-12-08
[github-action] build and upload Android APK as artifacts

wgtdkp:
This PR builds and uploads Android APK as artifacts.

See the failure: https://github.com/openthread/ot-commissioner/actions/runs/406484641

### DIFF
--- a/.github/workflows/android-app-release.yml
+++ b/.github/workflows/android-app-release.yml
@@ -25,8 +25,8 @@
 #  ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
 #  POSSIBILITY OF SUCH DAMAGE.
 #
-# This file runs Android App release build and publish
-# releases automatically.
+# This file runs Android App release build and uploads Android APK artifact for
+# creating releases.
 #
 
 name: Android App Release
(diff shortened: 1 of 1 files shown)


wgtdkp:
@jwhui The failure seems due to the `GITHUB_TOKEN` has only read permission for pull requests. [pull_request_target](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) seems to be the solution for such permission issues but I failed to get it work here.

For a quick fix, I switched to build the release and upload it as an artifact. After merging this PR, we can create a release with the [artifact](https://github.com/openthread/ot-commissioner/actions/runs/407764336). Does this sound good to you?

jwhui:
> @jwhui The failure seems due to the `GITHUB_TOKEN` has only read permission for pull requests. [pull_request_target](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull_request_target) seems to be the solution for such permission issues but I failed to get it work here.
> 
> For a quick fix, I switched to build the release and upload it as an artifact. After merging this PR, we can create a release with the [artifact](https://github.com/openthread/ot-commissioner/actions/runs/407764336). Does this sound good to you?

@wgtdkp , that works for me. Thanks!

jwhui (approved):

PR #170, opened 2021-03-05
[border-agent] add the discriminator TXT entry

wgtdkp:


src/app/border_agent.cpp
```diff
@@ -228,6 +228,21 @@ static int HandleRecord(const struct sockaddr *from,
             {
                 VerifyOrExit(value == "1", error = ERROR_BAD_FORMAT("value of TXT Key 'rv' is {} but not 1", value));
             }
+            else if (key == "dd")
+            {
+                auto &discriminator = binaryValue;
+
+                if (discriminator.size() != kExtendedAddrLength)
+                {
+                    ExitNow(error = ERROR_BAD_FORMAT("value of TXT Key 'dd' is invalid: value={}",
```
simonlingoogle, line 237:
```suggestion
                    borderAgent.mPresentFlags &= ~BorderAgent::kDiscriminatorBit;
                    ExitNow(error = ERROR_BAD_FORMAT("value of TXT Key 'dd' is invalid: value={}",
```

Necessary to mark it as not present?

simonlingoogle (approved):
LGTM :+1:

src/app/border_agent.cpp
wgtdkp: The `BorderAgent` default constructor initializes the flags to all zeros, so not necessary here.

jwhui (approved):

PR #177, opened 2021-03-19
[tests] add debug logging

wgtdkp:
Add more verbose test output to help debug when test cases fail.

### DIFF
--- a/tests/integration/common.sh
+++ b/tests/integration/common.sh
@@ -79,8 +79,6 @@ executable_or_die() {
 }
 
 start_daemon() {
-    set -e
-
     cd "${RUNTIME_DIR}"
 
     if pidof ot-daemon; then
@@ -94,8 +92,6 @@ start_daemon() {
 }
 
 stop_daemon() {
-    set -e
-
     sudo killall ot-daemon || true
     sleep 1
     (pidof ot-daemon && die "killing ot-daemon failed") || true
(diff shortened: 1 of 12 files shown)


bukepo (approved):

tests/integration/common.sh
```diff
@@ -220,15 +220,33 @@ form_network() {
     set -e
     local pskc=$1
 
+    echo "------ forming Thread network ------"
+
+    echo "sudo ${OT_CTL} channel ${CHANNEL}"
```
simonlingoogle, line 225: Being unfamiliar with the reason of adding these logging, should we use `set -x` instead ?

wgtdkp: changed to use `set -x`.

tests/integration/run_tests.sh
```diff
@@ -62,7 +64,7 @@ run_test_case() {
     ## we cannot declare output with `local`,
     ## because `local` is a command and its return value iwll override
     ## return value of the test case.
-    output=$(${test_case})
+    $(set -xeuo pipefail && ${test_case} &> "${output_file}")
```
wgtdkp, line 67: This is where we enables "xeuo pipefail" for all test cases.

simonlingoogle (approved):
LGTM :+1:

jwhui (approved):

PR #186, opened 2021-04-06
[dataset] fix and relax dataset response check

wgtdkp:
1. The Pending Dataset check is wrong: the bitwise or (`| `) should be bitwise and (`&`)
2. Both Pending and Active Dataset Get response checks are too restrict.

The Commissioner CLI always sends MGMT_ACTIVE/PENDING_GET.req for all TLVs even when only a single TLV is requested by the user and thus current implementation is working.


wgtdkp:
By the way, this is not related to the IOP issue with the Kirale BR.

jwhui (approved):

PR #188, opened 2021-04-11
[udp-proxy] track Mesh-Local prefix in UDP Proxy client

wgtdkp:
Previous implementation requires the caller of the Commissioner library to set Mesh-Local prefix before
initiating `MGMT` commands. it is error-prone and breaks backward compatibility: we need to first fetch
the Active Operational Dataset and set the Mesh-Local Prefix to the commissioner library before sending
other `MGMT_*` commands:
https://github.com/openthread/ot-commissioner/blob/1b5dd47875fdd52e86dd6334973c36fa9e1ba899/src/app/commissioner_app.cpp#L151-L155

This PR refactors the UDP Proxy to have it manage the Mesh-Local prefix inside the `ProxyClient` class.
1. The Mesh-Local prefix is lazily requested from the Border Agent  before sending the first `MGMT` command (if no Mesh-Local prefix is cached).
2. The cached Mesh-Local prefix is cleared when the commissioner receives `MGMT_DATASET_CHANGED.ntf` so that the latest Mesh-Local prefix will be fetched before next `MGMT` command.
3. The cached Mesh-Local prefix is cleared when the commissioner is disconnected.

In this way, the user doesn't need to be aware of the Mesh-Local prefix and no changes are required.

### DIFF
--- a/src/library/commissioner_impl.cpp
+++ b/src/library/commissioner_impl.cpp
@@ -48,9 +48,8 @@ namespace ot {
 
 namespace commissioner {
 
-static constexpr size_t kMeshLocalPrefixLength = 8;
-
-static constexpr uint16_t kLeaderAloc16 = 0xFC00;
+static constexpr uint16_t kLeaderAloc16     = 0xFC00;
+static constexpr uint16_t kPrimaryBbrAloc16 = 0xFC38;
 
 static constexpr uint16_t kDefaultMmPort = 61631;
 
@@ -128,7 +127,7 @@ CommissionerImpl::CommissionerImpl(CommissionerHandler &aHandler, struct event_b
     , mJoinerSessionTimer(mEventBase, [this](Timer &aTimer) { HandleJoinerSessionTimer(aTimer); })
     , mResourceUdpRx(uri::kUdpRx, [this](const coap::Request &aRequest) { mProxyClient.HandleUdpRx(aRequest); })
     , mResourceRlyRx(uri::kRelayRx, [this](const coap::Request &aRequest) { HandleRlyRx(aRequest); })
-    , mProxyClient(mEventBase, mBrClient)
+    , mProxyClient(*this, mBrClient)
 #if OT_COMM_CONFIG_CCM_ENABLE
     , mTokenManager(mEventBase)
 #endif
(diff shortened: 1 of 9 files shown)


src/library/commissioner_impl.cpp
```diff
@@ -495,15 +473,10 @@ void CommissionerImpl::GetRawActiveDataset(Handler<ByteArray> aHandler, uint16_t
...
-        // Request Active Dataset from the BA when we don't possess the mesh-local prefix.
-        mBrClient.SendRequest(request, onResponse);
-    }
-    else
-    {
-        mProxyClient.SendRequest(request, onResponse, GetLeaderLocator(), kDefaultMmPort);
-    }
+    // Always request Active Dataset from the Border Agent because
+    // the Mesh-Local prefix of current network is agnostic before
+    // we requesting the latest Active Operational Dataset.
```
jwhui, line 478: Can you try to reword this? I don't quite understand the "agnostic before we requesting" wording.

wgtdkp: Updated, please review.

src/library/udp_proxy.cpp
```diff
@@ -129,6 +162,61 @@ void ProxyClient::HandleUdpRx(const coap::Request &aUdpRx)
...
+Error ProxyClient::SetMeshLocalPrefix(const ByteArray &aMeshLocalPrefix)
+{
+    constexpr uint8_t meshLocalPrefix[] = {0xfd};
+    Error             error;
+
+    VerifyOrExit(aMeshLocalPrefix.size() == kMeshLocalPrefixLength,
+                 error = ERROR_INVALID_ARGS("Thread Mesh-local Prefix length must be {}", kMeshLocalPrefixLength));
+    VerifyOrExit(
+        std::equal(aMeshLocalPrefix.begin(), aMeshLocalPrefix.begin() + sizeof(meshLocalPrefix), meshLocalPrefix),
+        error = ERROR_INVALID_ARGS("Thread mesh-local Prefix must starts with fd00::/8"));
```
jwhui, line 193:
```suggestion
        error = ERROR_INVALID_ARGS("Thread Mesh-Local Prefix must start with fd00::/8"));
```

jwhui (approved):

PR #199, opened 2021-04-29
[api] add API for getting commissioner version

wgtdkp:
This commit adds a new commissioner `API` that returns the commissioner version.
Also add git revision to the version string.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,16 +57,16 @@ execute_process(
 )
 
 if(OT_COMM_GIT_REVISION)
-    set(OT_COMM_VERSION "${PROJECT_VERSION}-${OT_COMM_GIT_REVISION}")
+    set(OT_COMM_VERSION "${PROJECT_VERSION}+${OT_COMM_GIT_REVISION}")
 else()
     set(OT_COMM_VERSION "${PROJECT_VERSION}")
 endif()
 
 message(STATUS "Version: ${OT_COMM_VERSION}")
 
-add_library(commissioner-version INTERFACE)
-target_compile_definitions(commissioner-version
-    INTERFACE OT_COMM_VERSION="${PROJECT_VERSION}"
+add_library(commissioner-config INTERFACE)
+target_compile_definitions(commissioner-config
+    INTERFACE OT_COMM_VERSION="${OT_COMM_VERSION}"
 )
 
 add_subdirectory(src)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -27,7 +27,7 @@
 #
 
 cmake_minimum_required(VERSION 3.10.1)
-project(ot-commissioner VERSION 0.1.0)
+project(ot-commissioner VERSION 0.2.0)
 
 option(OT_COMM_ANDROID           "Build with Android NDK" OFF)
 option(OT_COMM_APP               "Build the CLI App" ON)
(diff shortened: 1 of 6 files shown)


include/commissioner/commissioner.hpp
```diff
@@ -1133,6 +1133,15 @@ class Commissioner
...
      */
     static void AddJoiner(ByteArray &aSteeringData, const ByteArray &aJoinerId);
+
+    /**
+     * @brief Return the commissioner version.
+     *
+     * @return A version string in format of <MAJOR>.<MINOR>.<PATCH>[-<GIT_REVISION>].
+     *         The GIT_REVISION is included only when the commissioner is built in a
+     *         git repository.
+     */
```
simonlingoogle, line 1143: Better not include git revision to make the format consistent? 
Maybe we can add another api to get the git revision if necessary.

wgtdkp: Actually, I want the git revision more than the semantic version numbers. It is the usual situation that we want the exact commit ID when there are IOP issues. I believe OpenThread and ot-br-posix are already including the git revision in the version string and we are just following it here.

What is different here is that we provide an API for getting the version string but not defining a macro or use auto-generated version file. This is because ot-commissioner can be built and installed to system directory and used in a project that is not using CMake or automake tools.

simonlingoogle: I rechecked https://semver.org/#spec-item-10. 

Seems we need to use the format:
`<MAJOR>.<MINOR>.<PATCH>[+<GIT_REVISION>]` because `GIT_REVISION` should be `Build metadata`.

include/commissioner/commissioner.hpp
```diff
@@ -1133,6 +1133,15 @@ class Commissioner
      * @param[in]      aJoinerId      A Joiner ID.
      */
     static void AddJoiner(ByteArray &aSteeringData, const ByteArray &aJoinerId);
+
+    /**
+     * @brief Return the commissioner version.
+     *
+     * @return A version string in format of <MAJOR>.<MINOR>.<PATCH>[-<GIT_REVISION>].
```
simonlingoogle, line 1140:
```suggestion
     * @return A version string in format of <MAJOR>.<MINOR>.<PATCH>[+<GIT_REVISION>].
```
inline with https://semver.org/#spec-item-10

CMakeLists.txt
```diff
@@ -50,6 +50,25 @@ endif()
...
 set(CMAKE_CXX_EXTENSIONS OFF)
 
+execute_process(
+    COMMAND git describe --dirty --always 2>/dev/null
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    OUTPUT_VARIABLE OT_COMM_GIT_REVISION OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+if(OT_COMM_GIT_REVISION)
+    set(OT_COMM_VERSION "${PROJECT_VERSION}-${OT_COMM_GIT_REVISION}")
```
simonlingoogle, line 60:
```suggestion
    set(OT_COMM_VERSION "${PROJECT_VERSION}+${OT_COMM_GIT_REVISION}")
```

CMakeLists.txt
```diff
@@ -50,6 +50,25 @@ endif()
...
+    COMMAND git describe --dirty --always 2>/dev/null
+    WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+    OUTPUT_VARIABLE OT_COMM_GIT_REVISION OUTPUT_STRIP_TRAILING_WHITESPACE
+)
+
+if(OT_COMM_GIT_REVISION)
+    set(OT_COMM_VERSION "${PROJECT_VERSION}+${OT_COMM_GIT_REVISION}")
+else()
+    set(OT_COMM_VERSION "${PROJECT_VERSION}")
+endif()
```
bukepo, line 63: Allow customize version.
```suggestion
if(NOT OT_COMM_VERSION)
    if(OT_COMM_GIT_REVISION)
        set(OT_COMM_VERSION "${PROJECT_VERSION}+${OT_COMM_GIT_REVISION}")
    else()
        set(OT_COMM_VERSION "${PROJECT_VERSION}")
    endif()
endif()
```

CMakeLists.txt
```diff
@@ -50,6 +50,25 @@ endif()
...
+
+message(STATUS "Version: ${OT_COMM_VERSION}")
+
+add_library(commissioner-version INTERFACE)
```
bukepo, line 67: For future flags?
```suggestion
add_library(ot-comm-config INTERFACE)
```

CMakeLists.txt
wgtdkp: Are there cases that we want need to specify a version other that the project version? I think there is no other source of the version.

CMakeLists.txt
wgtdkp: renamed to `commissioner-config` to follow existing targets naming.

simonlingoogle (approved):
LGTM :+1:

bukepo (approved):

jwhui (approved):

PR #200, opened 2021-04-30
[script] add build script

bukepo:


wgtdkp (commented):
@bukepo you probably want to avoid switching to the build directory for building, right? For this purpose, you can use below cmake commands to config, build and install cmake projects from project directory:

```shell
cmake -S . -B build -GNinja
cmake --build build
cmake --install build

```

bukepo:
> @bukepo you probably want to avoid switching to the build directory for building, right? For this purpose, you can use below cmake commands to config, build and install cmake projects from project directory:
> 
> ```shell
> cmake -S . -B build -GNinja
> cmake --build build
> cmake --install build
> ```

just to keep consistent with other OpenThread derived projects.

.gitignore
```diff
@@ -45,7 +45,7 @@ Makefile.in
 
 # build
 .build*
-build
+/build
```
wgtdkp, line 48: What's the difference here?

wgtdkp (approved):

.gitignore
bukepo: `build` ignores all files named `build` under sub directories.

wgtdkp: Isn't it what we want?

wgtdkp: Probably you want it be `build/` to ignore only directories but not files?

bukepo: Actually I see many projects(e.g. CHIP) have a build directory tracked. I think we can only make such assumption for the current directory.

PR #202, opened 2021-05-06
OT Commissioner enhancements

mikhail-pridushchenko:

PR #203, opened 2021-05-29
[cli] support default config without specifying a config file

wgtdkp:
This PR includes two small enhancements:
1. Support starting the `commissioner-cli` without specifying a configuration file.
    a. pskc is set to all-one.
    b. log messages are printed to syslog (/var/log/syslog).
2. Add sys logger support.

### DIFF
--- a/src/app/cli/main.cpp
+++ b/src/app/cli/main.cpp
@@ -63,7 +63,7 @@ static void PrintUsage(const std::string &aProgram)
 {
     static const std::string usage = "usage: \n"
                                      "    " +
-                                     aProgram + " <config-file>";
+                                     aProgram + " [<config-file>]";
 
     Console::Write(usage, Console::Color::kWhite);
 }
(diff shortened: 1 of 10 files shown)


src/app/cli/main.cpp
```diff
@@ -89,18 +89,25 @@ static void HandleSignalInterrupt()
 
 int main(int argc, const char *argv[])
 {
-    Error  error;
-    Config config;
+    Error       error;
```
bukepo, line 92: may enhance to use getopt in future PRs.

bukepo (approved):
Thanks! :+1:
Is it possible to add a test or update existing tests without specifying the config file?

src/app/cli/main.cpp
wgtdkp: I see DSR's PR https://github.com/openthread/ot-commissioner/pull/202 already changed to use `getopt` so let that PR do the change.

tests/integration/test_joining.sh
```diff
@@ -47,6 +47,26 @@ test_joining() {
...
+    ## enable all MeshCoP joiners
+    send_command_to_commissioner "joiner enable meshcop ${JOINER_EUI64} ${JOINER_CREDENTIAL}"
+
+    start_joiner "meshcop"
+
+    stop_commissioner
+
+    stop_daemon
+}
+
```
wgtdkp, line 69: @bukepo The test case for starting without a config file.

jwhui (approved):

PR #211, opened 2021-06-18
[cli] fix "borderagent" command documentation

wgtdkp:
This commit adds the CLI documentation of the `borderagent discover` command.
Also fixes a bug for the `borderagent get locator` command that the locator is not
pulled to the commissioner.


EskoDijk (approved):
Thanks! Looks clear now.

jwhui (approved):

PR #212, opened 2021-06-21
[dataset] avoid setting session ID or Border Agent Locator

wgtdkp:
`Commissioner Session ID` and `Border Agent Locator` in the `Commissioner Dataset` are read-only.
Current implementation rejects setting the Commissioner Dataset if any of the fields is present, but it
is not convenient for the caller to manually clear out those fields. This commits change the implementation
to always clear out the `Commissioner Session ID` and `Border Agent Locator` fields before sending the
`MGMT_COMM_SET.req` message to the border agent.

Fixes master CI failure https://github.com/openthread/ot-commissioner/runs/2861632149.


simonlingoogle (approved):
LGTM :+1:

jwhui (approved):
Thanks! 👍🏼

PR #228, opened 2022-04-08
[test] increase wait time for leader start

abtink:
This commit increases the wait time for leader start. This is to
address the changes in behavior in Thread spec related to the
number of MLE Parent Request messages that device needs to send in
the first attach attempt before it can become leader.  The new design
requires a device to send a total of six Parent Request messages,
first two to routers followed by four to routers and REEDs, each
having its own corresponding wait time.


wgtdkp (approved):
Thanks @abtink !

jwhui (approved):

PR #229, opened 2022-06-27
[test] update leader start in `test_native_commissioner()`

abtink:
This commit increases the wait time for leader start. This is to
address the changes in behavior in Thread spec related to the
number of MLE Parent Request messages that device needs to send in
the first attach attempt before it can become leader.  The new design
requires a device to send a total of six Parent Request messages,
first two to routers followed by four to routers and REEDs, each
having its own corresponding wait time.

### DIFF
--- a/tests/integration/test_native_commissioner.sh
+++ b/tests/integration/test_native_commissioner.sh
@@ -44,7 +44,7 @@ send "pskc ${PSKC}\r\n"
 expect "Done"
 send "thread start\r\n"
 expect "Done"
-sleep 3
+sleep 10
 send "state\r\n"
 expect "leader"
 expect "Done"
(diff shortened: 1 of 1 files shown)


wgtdkp (approved):
Thanks!

abtink:
The CI tests are failing with this change. I re-ran it a bunch of times but it keeps failing.
https://github.com/openthread/ot-commissioner/runs/7081574749?check_suite_focus=true

wgtdkp:
> The CI tests are failing with this change. I re-ran it a bunch of times but it keeps failing.
> https://github.com/openthread/ot-commissioner/runs/7081574749?check_suite_focus=true
> 

Sorry I missed the message here, looks like you already figured out the fix!

abtink:
> Sorry I missed the message here, looks like you already figured out the fix!

No worries. Thanks. Yes. We needed to increase the `sleep` delay after calling the `start_leader` as well 😄

jwhui (approved):

PR #249, opened 2024-03-31
[build] support mbedtls 3.x

wgtdkp:
This PR adds support for building with mbedtls 3.x. Specifically:
1. some mbedtls APIs are removed or renamed in 3.x so that we need to do the migration
2. fix some fmtlib compile issues, e.g. making enums formattable 
3. we changed to use MBEDTLS_CONFIG_FILE to specify the full configuration instead of leveraging MBEDTLS_USER_CONFIG_FILE on top of the default mbedtls configuration (this is for working around internal building issues)


librasungirl (approved):
👍

jwhui (approved):

PR #250, opened 2024-04-01
[build] fix issue when executing `build-commissioner-libs.sh` multiple times

ZhangLe2016:
When developers run the 'build-commissioner-libs.sh' script multiple times, they might see the following error during compiling the libevent:

error: call to undeclared function 'pipe2'; ISO C99 and later do not support implicit function declarations ...

In fact, we did not use pip2 in libevent. We can turn it off by using the macro EVENT_HAVE-PIPE2=0


ZhangLe2016:
@jwhui seems that I cannot add reviewers into this PR, could you please add me into this project as a contributor? thanks.

wgtdkp:
Better add this to third_party/libevent/CMakeLists.txt:

```
set(EVENT__HAVE_PIPE2 0 CACHE BOOL "disable use of pipe2 in libevent" FORCE)
```

android/build-commissioner-libs.sh
```diff
@@ -55,6 +55,7 @@ cmake -GNinja \
     -DCMAKE_CXX_STANDARD=11 \
     -DCMAKE_CXX_STANDARD_REQUIRED=ON \
     -DCMAKE_BUILD_TYPE=Release \
+    -DEVENT__HAVE_PIPE2=0 \
```
wgtdkp, line 58: This can be removed now

ZhangLe2016: Oops! Thanks for the reminder.

wgtdkp (approved):
LGTM, Thanks! :+1:

librasungirl (approved):
👍

jwhui (approved):

PR #251, opened 2024-04-02
[android] optimize image size with optimage

wgtdkp:
Use https://github.com/sk-/optimage to perform lossless image size optimization to make the app smaller


librasungirl (approved):
👍

jwhui (approved):

PR #256, opened 2024-04-10
[cli] support Thread Administrator Passcode configuration

librasungirl:


src/app/cli/interpreter.cpp
```diff
@@ -912,17 +914,37 @@ Interpreter::Value Interpreter::ProcessConfig(const Expression &aExpr)
...
+        if (aExpr[2] == "accesscode")
+        {
+            mThreadAccessPasscode = aExpr[3];
+
+            // TODO(rongli@): add len and last digit check per SPEC-1216v11
+
+            for (char c : mThreadAccessPasscode)
+            {
+                pskc.push_back(static_cast<uint8_t>(c));
+            }
```
wgtdkp, line 942: nit:
```suggestion
            pskc.assign(mThreadAccessCode.begin(), mThreadAccessCode.end());
```

wgtdkp (approved):
Thanks, LGTM!

src/app/cli/interpreter.cpp
```diff
@@ -197,7 +197,9 @@ const std::map<std::string, Interpreter::Evaluator> &Interpreter::mEvaluatorMap
 };
 
 const std::map<std::string, std::string> &Interpreter::mUsageMap = *new std::map<std::string, std::string>{
-    {"config", "config get pskc\n"
+    {"config", "config get accesscode\n"
+               "config set accesscode <9-digits>\n"
```
lanyuwen, line 201: not sure if I have some misunderstanding here, but get/set accesscode seems somewhat confusing to me, for which people might think it to get/set accesscode of border router. What about adding it as a parameter to get epskc? e.g., config get epskc -c <9-digits accesscode>?

librasungirl: updated with more friendly hints, pls take a look at the new patch

lanyuwen (approved):
LGTM

wgtdkp (approved):

src/app/cli/interpreter.cpp
```diff
@@ -197,7 +197,9 @@ const std::map<std::string, Interpreter::Evaluator> &Interpreter::mEvaluatorMap
 };
 
 const std::map<std::string, std::string> &Interpreter::mUsageMap = *new std::map<std::string, std::string>{
-    {"config", "config get pskc\n"
+    {"config", "config get accesscode\n"
+               "config set accesscode <9-digits-thread-access-passcode>\n"
```
wgtdkp, line 201: Just come to realize that if the access code has the empty spaces removed, then the string is effectively a epskc. So probably we should change this command to "config set epskc"?

librasungirl: Offline chat with @wgtdkp, fine to keep it consistent with the terminology with Thread Specification.

src/app/cli/interpreter.cpp
```diff
@@ -197,7 +197,9 @@ const std::map<std::string, Interpreter::Evaluator> &Interpreter::mEvaluatorMap
 };
 
 const std::map<std::string, std::string> &Interpreter::mUsageMap = *new std::map<std::string, std::string>{
-    {"config", "config get pskc\n"
+    {"config", "config get accesscode\n"
```
jwhui, line 200: nit: Thread Group TC call yesterday decided to call this "Thread Administrator Passcode".

librasungirl: rename in new patch

jwhui (approved):

PR #260, opened 2024-04-12
[cli] update `br scan` to allow network interface selection for mDNS binding

ZhangLe2016:
Within the OTBR practice application, the BR host will have several interfaces. When users want to discover the border router using mDNS, they will need to choose the specific interface.

This change introduces a new option for the CLI command `br scan`. User can now specify a network interface using the syntax `br scan --netif <network interface>. The chosen interface will be used for mDNS binding through socket options.

### DIFF
--- a/include/commissioner/error.hpp
+++ b/include/commissioner/error.hpp
@@ -161,7 +161,6 @@ enum class ErrorCode : int
      * The error is out of the address space of OT Commissioner.
      */
     kUnknown = 19,
-
 };
 
 /**
--- a/include/commissioner/error.hpp
+++ b/include/commissioner/error.hpp
@@ -161,6 +161,11 @@ enum class ErrorCode : int
      * The error is out of the address space of OT Commissioner.
      */
     kUnknown = 19,
+
+    /**
+     * The error occurs when attempting to bind a network interface to an mDNS socket."
+     */
+    kSocketBindError = 20,
 };
 
 /**
--- a/include/commissioner/error.hpp
+++ b/include/commissioner/error.hpp
@@ -162,10 +162,6 @@ enum class ErrorCode : int
      */
     kUnknown = 19,
 
-    /**
-     * The error occurs when attempting to bind a network interface to an mDNS socket."
-     */
-    kSocketBindError = 20,
 };
 
 /**
--- a/src/app/cli/interpreter.cpp
+++ b/src/app/cli/interpreter.cpp
@@ -1466,7 +1466,7 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
 
         if (!netIf.empty() && setsockopt(mdnsSocket, SOL_SOCKET, SO_BINDTODEVICE, netIf.c_str(), netIf.size()) < 0)
         {
-            ExitNow(value = ERROR_SOCKET_BIND_ERROR("failed to bind network interface: {}: {}", netIf, strerror(errno)));
+            ExitNow(value = ERROR_SOCKET_BIND_ERROR("failed to bind network interface {}: {}", netIf, strerror(errno)));
         }
 
         fdgMdnsSocket.mFD = mdnsSocket;
--- a/src/app/cli/interpreter.cpp
+++ b/src/app/cli/interpreter.cpp
@@ -46,6 +46,7 @@
 
 #include <fcntl.h>
 #include <sys/types.h>
+#include <sys/socket.h>
 #include <unistd.h>
 
 #include "app/border_agent.hpp"
--- a/src/app/cli/interpreter.cpp
+++ b/src/app/cli/interpreter.cpp
@@ -1443,13 +1443,7 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
             {
                 if (++it != mContext.mCommandKeys.end())
                 {
-                    try
-                    {
-                        scanTimeout = stol(*it);
-                    } catch (...)
-                    {
-                        ExitNow(value = ERROR_INVALID_ARGS("Imparsable timeout value '{}'", *it));
-                    }
+                    SuccessOrExit(value = ParseInteger(scanTimeout, *it));
                 }
                 else
                 {
--- a/src/app/cli/interpreter.cpp
+++ b/src/app/cli/interpreter.cpp
@@ -1466,7 +1466,7 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
 
         if (!netIf.empty() && setsockopt(mdnsSocket, SOL_SOCKET, SO_BINDTODEVICE, netIf.c_str(), netIf.size()) < 0)
         {
-            ExitNow(value = ERROR_SOCKET_BIND_ERROR("failed to bind network interface: {}", netIf));
+            ExitNow(value = ERROR_SOCKET_BIND_ERROR("failed to bind network interface: {}: {}", netIf, strerror(errno)));
         }
 
         fdgMdnsSocket.mFD = mdnsSocket;
--- a/src/app/cli/interpreter.cpp
+++ b/src/app/cli/interpreter.cpp
@@ -1437,29 +1437,29 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
         nlohmann::json                            baJson;
         char                                      mdnsSendBuffer[kMdnsBufferSize];
 
-        for (auto it = mContext.mCommandKeys.begin(); it != mContext.mCommandKeys.end(); ++it)
+        auto it = std::find(mContext.mCommandKeys.begin(), mContext.mCommandKeys.end(), "--timeout");
+        if (it != mContext.mCommandKeys.end())
         {
-            if (*it == "--timeout")
+            if (++it != mContext.mCommandKeys.end())
             {
-                if (++it != mContext.mCommandKeys.end())
-                {
-                    SuccessOrExit(value = ParseInteger(scanTimeout, *it));
-                }
-                else
-                {
-                    ExitNow(value = ERROR_INVALID_ARGS("Missing {} value", *--it));
-                }
+                SuccessOrExit(value = ParseInteger(scanTimeout, it[0]));
             }
-            else if (*it == "--netif")
+            else
             {
-                if (++it != mContext.mCommandKeys.end())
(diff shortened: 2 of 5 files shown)


src/app/cli/interpreter.cpp
```diff
@@ -1443,9 +1444,19 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
...
+        if (mContext.mCommandKeys.size() == 2 && mContext.mCommandKeys[0] == "--netif")
+        {
+            netIf = mContext.mCommandKeys[1];
+        }
+
         // Open IPv4 mDNS socket
         mdnsSocket = mdns_socket_open_ipv4();
         VerifyOrExit(mdnsSocket >= 0, value = ERROR_IO_ERROR("failed to open mDNS IPv4 socket"));
+
+        if (netIf != "" && setsockopt(mdnsSocket, SOL_SOCKET, SO_BINDTODEVICE, &netIf[0], sizeof(netIf)) < 0) {
```
wgtdkp, line 1456:
```suggestion
        if (!netIf.empty() && setsockopt(mdnsSocket, SOL_SOCKET, SO_BINDTODEVICE, netIf.c_str(), netIf.size()) < 0) {
```

src/app/cli/interpreter.cpp
```diff
@@ -1443,9 +1444,19 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
             }
         }
 
+        if (mContext.mCommandKeys.size() == 2 && mContext.mCommandKeys[0] == "--netif")
```
wgtdkp, line 1447: Can we support multiple options e.g. "br scan --timeout 400 --netif wlan0"?

src/app/cli/interpreter.cpp
```diff
@@ -1443,9 +1444,19 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
...
+        {
+            netIf = mContext.mCommandKeys[1];
+        }
+
         // Open IPv4 mDNS socket
         mdnsSocket = mdns_socket_open_ipv4();
         VerifyOrExit(mdnsSocket >= 0, value = ERROR_IO_ERROR("failed to open mDNS IPv4 socket"));
+
+        if (netIf != "" && setsockopt(mdnsSocket, SOL_SOCKET, SO_BINDTODEVICE, &netIf[0], sizeof(netIf)) < 0) {
+          ExitNow(value = ERROR_IO_ERROR("failed to bind network interface: {}", netIf));
```
wgtdkp, line 1457: Please add the error code to the returned value so that the user knows what's wrong here

src/app/cli/interpreter.cpp
ZhangLe2016: Modified this part to support both options.

src/app/cli/interpreter.cpp
```diff
@@ -1432,20 +1433,32 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
...
+        if (it != mContext.mCommandKeys.end()) {
+            netIf = mContext.mCommandKeys[std::distance(mContext.mCommandKeys.begin(), it) + 1];
+        }
+
         // Open IPv4 mDNS socket
         mdnsSocket = mdns_socket_open_ipv4();
         VerifyOrExit(mdnsSocket >= 0, value = ERROR_IO_ERROR("failed to open mDNS IPv4 socket"));
+
+        if (!netIf.empty() && setsockopt(mdnsSocket, SOL_SOCKET, SO_BINDTODEVICE, netIf.c_str(), netIf.size()) < 0) {
+            ExitNow(value = ERROR_SOCKET_BIND_ERROR("failed to bind network interface: {}", netIf));
```
wgtdkp, line 1462:
```suggestion
            ExitNow(value = ERROR_SOCKET_BIND_ERROR("failed to bind network interface {}: {}", netIf, strerror(errno)));
```

src/app/cli/interpreter.cpp
```diff
@@ -1432,20 +1433,32 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
...
 
-        if (mContext.mCommandKeys.size() == 2 && mContext.mCommandKeys[0] == "--timeout")
-        {
+        auto it = std::find(mContext.mCommandKeys.begin(), mContext.mCommandKeys.end(), "--timeout");
+
+        if (it != mContext.mCommandKeys.end()) {
             try
             {
-                scanTimeout = stol(mContext.mCommandKeys[1]);
+              scanTimeout = stol(mContext.mCommandKeys[std::distance(mContext.mCommandKeys.begin(), it) + 1]);
```
wgtdkp, line 1442: I think you can use `it[1]`:
```suggestion
              scanTimeout = stol(it[1]);
```

src/app/cli/interpreter.cpp
```diff
@@ -1432,20 +1433,32 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
         nlohmann::json                            baJson;
         char                                      mdnsSendBuffer[kMdnsBufferSize];
 
-        if (mContext.mCommandKeys.size() == 2 && mContext.mCommandKeys[0] == "--timeout")
-        {
+        auto it = std::find(mContext.mCommandKeys.begin(), mContext.mCommandKeys.end(), "--timeout");
+
+        if (it != mContext.mCommandKeys.end()) {
```
wgtdkp: This will crash if the user provides only "--timeout", e.g. "br scan --timeout".
```suggestion
        if (it != mContext.mCommandKeys.end() && it +1 != mContext.mCommandKeys.end()) {
```

wgtdkp: The same for the "--netif" option

include/commissioner/error.hpp
```diff
@@ -161,6 +161,11 @@ enum class ErrorCode : int
      * The error is out of the address space of OT Commissioner.
      */
     kUnknown = 19,
+
+    /**
+     * The error occurs when attempting to bind a network interface to an mDNS socket."
+     */
+    kSocketBindError = 20,
```
wgtdkp, line 168: We don't need this new error code, right?

src/app/cli/interpreter.cpp
```diff
@@ -1432,20 +1437,47 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
...
             }
         }
 
         // Open IPv4 mDNS socket
         mdnsSocket = mdns_socket_open_ipv4();
         VerifyOrExit(mdnsSocket >= 0, value = ERROR_IO_ERROR("failed to open mDNS IPv4 socket"));
+
+        if (!netIf.empty() && setsockopt(mdnsSocket, SOL_SOCKET, SO_BINDTODEVICE, netIf.c_str(), netIf.size()) < 0)
+        {
+            ExitNow(value = ERROR_SOCKET_BIND_ERROR("failed to bind network interface {}: {}", netIf, strerror(errno)));
```
wgtdkp, line 1478:
```suggestion
            ExitNow(value = ERROR_INVALID_ARGS("failed to bind network interface {}: {}", netIf, strerror(errno)));
```

src/app/cli/interpreter.cpp
```diff
@@ -1432,20 +1437,47 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
...
+                if (++it != mContext.mCommandKeys.end())
+                {
+                    netIf = *it;
+                }
+                else
+                {
+                    ExitNow(value = ERROR_INVALID_ARGS("Missing {} value", *--it));
+                }
             }
         }
```
wgtdkp, line 1470: I prefer your previous implementation which use `std::find` to get the options you care about, which is more clearer and straight forward :)

The current implementation here needs to be very careful with the iterators (the ++, --) and I believe you spent some time to get it correct. So it's really harder to maintain this version.

Please consider using the `std::find` version.

src/app/cli/interpreter.cpp
```diff
@@ -1432,20 +1437,47 @@ Interpreter::Value Interpreter::ProcessBr(const Expression &aExpr)
...
-            } catch (...)
+                if (++it != mContext.mCommandKeys.end())
+                {
+                    try
+                    {
+                        scanTimeout = stol(*it);
+                    } catch (...)
+                    {
+                        ExitNow(value = ERROR_INVALID_ARGS("Imparsable timeout value '{}'", *it));
+                    }
```
wgtdkp, line 1452: Basically, try to avoid Exception in OT Commissioner code. As an alternative, you can use the c method `strtol()`.
 Additionally, better add a helper method like
```
Error ParseInt(const char* aStr, int &aValue);
```

src/app/cli/interpreter.cpp
ZhangLe2016: I agree that `std::find` is easier to understand and use. However, the current method might be more efficient. It only iterates through `mContext.mCommandKeys` once, whereas `std::find` might need to iterate multiple times.

src/app/cli/interpreter.cpp
ZhangLe2016: I found we have a dedicated helper method named `emplate <typename T> static Error ParseInteger(T &aInteger, const std::string &aStr)` which could be used here to replace the `Exception` block.

src/app/cli/interpreter.cpp
wgtdkp: Performance doesn't matter in a CLI tool like this :)

ZhangLe2016: Okay, Let me update it again

ZhangLe2016: updated, Thanks.

include/commissioner/error.hpp
```diff
@@ -161,6 +161,7 @@ enum class ErrorCode : int
      * The error is out of the address space of OT Commissioner.
      */
     kUnknown = 19,
+
```
wgtdkp, line 164: nit: revert this blank line

wgtdkp (approved):
:+1: 
Thanks!!

include/commissioner/error.hpp
wgtdkp, line 164:
```suggestion
```

jwhui (approved):

PR #262, opened 2024-04-15
[build] update cmake to 3.20.1

wgtdkp:
Updates cmake to 3.20.1 which adds support for USE_SWIG_DEPENDENCIES

USE_SWIG_DEPENDENCIES is needed to tell SWIG compiler to rebuild the java
library when the dependencies are changed. For example, we currently have a problem
that the java library is not updated even a new API is added to commissioner.hpp


librasungirl (approved):
👍

jwhui (approved):

PR #263, opened 2024-04-25
[cli] remove definition of `SO_BINDTODEVICE` from interpreter module

ZhangLe2016:
A previously introduced macro `SO_BINDOTDEVICE`, intended to pass the pre-commit code checks, which is conflicted with an definition in header file named <sys/socket.h>. So we need remove the macro from the interpreter module. To ensure the pre-commit code checks pass, we need to include the approriate header file and call `setsocketopt` with OS-specific parameters.

### DIFF
--- a/src/app/cli/interpreter.cpp
+++ b/src/app/cli/interpreter.cpp
@@ -45,7 +45,11 @@
 #include <vector>
 
 #include <fcntl.h>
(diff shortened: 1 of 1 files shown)


wgtdkp (approved):
LGTM :+1:

jwhui (approved):

PR #266, opened 2024-05-07
[tlv] add new tlvs for network diagnostic

ZhangLe2016:
libotcommissioner.jar, a versatile Java library, provides support for wide range of TMF message transactions, including MeshCop protocols and diagnostic TLVs. This PR focuses on integrating all the diagnostic TLVs, as defined in the spec `10.11.3` into the library's TLV module


src/library/tlv.cpp
```diff
@@ -220,6 +220,71 @@ bool Tlv::IsValid() const
     {
         return false;
     }
+    else if (mScope == Scope::kNetworkDiag)
+    {
+        switch (mType)
+        {
+        // Network disgnostic layer TLVs
+        case Type::kNetworkDiagExtMacAddress:
+            return length == 8;
```
abtink, line 229: We generally consider a TLV valid if its length is at least as long as the currently defined value length. This  accommodates potential future extensions of the TLV format.

ZhangLe2016: @abtink Thank you for your review. I have clarified the length of TLVs as per the definition in the spec. For instance, the length of the extended MAC address TLV in spec 5.19.2 is 8. Could you please kindly indicate any other inappropriate definitions in this PR?

wgtdkp: I think @abtink wants you to change the check to `return length >= 8;` so that the extended MAC extended address can be extended to maybe 16 bytes in Thread 2.0 or whatever later versions

ZhangLe2016: got it, Thanks @wgtdkp clarification. Let me update check condition.

wgtdkp (approved):

lanyuwen (approved):
LGTM

src/library/tlv.cpp
```diff
@@ -220,6 +220,73 @@ bool Tlv::IsValid() const
...
+        case Type::kNetworkDiagVersion:
+            return length == 2;
+        case Type::kNetworkDiagVendorName:
+            return length <= 4;
+        case Type::kNetworkDiagVendorModel:
+            return length <= 4;
+        case Type::kNetworkDiagVendorSWVersion:
+            return length <= 2;
+        case Type::kNetworkDiagChild:
+            return length <= 43;
```
abtink, line 275: Should these also use `>=`?

ZhangLe2016: As I understand it, the TLV will only have content when the queried device has children. If the queried device has no children, the value of TLV is empty, so the length should be 0. If we check the length to be `>=` 43, then when the value of TLV is empty, the TLV will be considered invalid.

PR #269, opened 2024-05-07
[diag] add a function to generate diag type list TLV based on a flag

ZhangLe2016:
Based on spec `10.11.4.6`, the type list TLV(18) concatenates the list of type identifiers of diagnostic TLVs. This TLV is used to request or reset the diag vlaues from other Thread devices.

This PR implements a function to generate the type list TLV vlaue based on a flag. Users can custmize the flag based on the diag TLVs type value. E.x. if users want to requeset or reset type 0, 1, 2 TLVs, they can set the flag to 0x7, where the position of bits 111 represent the required TLV types.

### DIFF
--- a/src/library/commissioner_impl.cpp
+++ b/src/library/commissioner_impl.cpp
@@ -1901,6 +1901,129 @@ exit:
     return error;
 }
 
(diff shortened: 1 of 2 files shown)


ZhangLe2016:
Close this PR for merging it into #270

PR #270, opened 2024-05-07
[api] enable DIAG_GET request command into commissioner

ZhangLe2016:
integrate and implement the DIAG_GET reqest command into commissioner module


include/commissioner/commissioner.hpp
```diff
@@ -1146,6 +1146,30 @@ class Commissioner
...
+     * @brief Asynchronously command a Thread device to request a set of diagnostic TLV(s) data.
+     *
+     * This method requests a Thread device to request a set of diagnostic data by sending DIAG_GET.req message.
+     * It always returns immediately without waiting for the completion. //TBC by real test
+     *
+     * @param[in, out] aHandler       A handler of all response and errors; Guaranteed to be called.
+     * @param[in]      aRloc          The RLOC of dest Thread device.
+     * @param[in]      aDiagTlvFlags  Diagnostic TLVs flags indicate which TLVs are wanted.
+     */
+    virtual void CommandDiagGet(Handler<ByteArray> aHandler, uint16_t aRloc, uint64_t aaDiagTlvFlags) = 0;
```
wgtdkp, line 1160: How can an app get the Rloc of a Thread device? Is the IP address easier to retrieved?

@librasungirl thoughts?

librasungirl: Depends on the use cases, there are diag TLVs (child table tlv, child tlv, child ipv6 address list tlv, router nerighbor tlv) which incluce rloc16.

RLOC (RLOC16) is routing locator in mesh, though it has the benefits of not raising address query which is realm local multicast, it will probably change, more chance in larger scale network. Instead MLEID is quite reliable, comparing with RLOC or OMR (off-mesh routable prefix may change when the BR offering OMR prefix changes), and will be preferred if periodically query the specific TLV lists from the specific device.

wgtdkp: It sounds like it more extensible and stable to use an IP address here rather than a Rloc?

ZhangLe2016:
Let me close this PR and resubmit a new one for this feature.

PR #274, opened 2024-05-07
[api] add a function to check the diagnostic TLV type is intended for query

ZhangLe2016:


src/library/commissioner_impl.hpp
```diff
@@ -251,6 +251,17 @@ class CommissionerImpl : public Commissioner
 
     void HandleJoinerSessionTimer(Timer &aTimer);
 
+    bool IsDiagQueryId(uint16_t aQueryId) const
```
wgtdkp, line 254: Can this change be merged into a PR which actually use it? We don't need to split PRs small like this. For example, it's actually harder to review this PR without knowing how this method will be used.

src/library/commissioner_impl.hpp
```diff
@@ -251,6 +251,17 @@ class CommissionerImpl : public Commissioner
...
+    {
+        if ((aQueryId == (uint16_t)tlv::Type::kNetworkDiagChild) ||
+            (aQueryId == (uint16_t)tlv::Type::kNetworkDiagChildIpv6Address) ||
+            (aQueryId == (uint16_t)tlv::Type::kNetworkDiagRouterNeighbor))
+        {
+            return true;
+        }
+        return false;
```
wgtdkp, line 262:
```suggestion
        return ((aQueryId == (uint16_t)tlv::Type::kNetworkDiagChild) ||
            (aQueryId == (uint16_t)tlv::Type::kNetworkDiagChildIpv6Address) ||
            (aQueryId == (uint16_t)tlv::Type::kNetworkDiagRouterNeighbor));
```

src/library/commissioner_impl.hpp
ZhangLe2016: make sense, Let me cancel it and merge it to #275. Thanks.

PR #278, opened 2024-05-13
[coap] fix the code to ack the confirmable coap response rather than respond with the payload over the dtls session

ZhangLe2016:
based on the CoAP protocol, when a CoAP server receives a confirmable CoAP response from another device, it should send a CoAP ACK message. It is incorrect to directly respond with the response data through the DTLS session.

### DIFF
--- a/src/library/coap.cpp
+++ b/src/library/coap.cpp
@@ -653,7 +653,7 @@ void Coap::HandleRequest(const Request &aRequest)
     {
         LOG_INFO(LOG_REGION_COAP, "server(={}) found cached CoAP response for resource {}", static_cast<void *>(this),
                  uriPath);
-        ExitNow(error = Send(*response));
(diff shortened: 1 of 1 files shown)


wgtdkp:
Could you share a link to the specific section in the CoAP RFC? I can take closer look.

ZhangLe2016:
> Could you share a link to the specific section in the CoAP RFC? I can take closer look.

Hi @wgtdkp thanks a lot for your reviewing, I checked `[5.2.2](https://datatracker.ietf.org/doc/html/rfc7252#section-5.2.2).  Separate` this part.
image

ZhangLe2016:
Discussed with @wgtdkp, there is no problem on the current implementation of CoAP, I need to handle request messages from application layer.  So let me close this PR.

PR #279, opened 2024-05-22
[udp-proxy] set local port to dedicated port number(:mm) for udp proxy transaction

ZhangLe2016:
Currently, while the commissioner is transmitting udp_tx messages, the UDP source port is provided by the `DtlsSession` interface. This value is randomly generated. However, the CoAP layer on the border router side performs checks on the received udp_tx port to ensure that the destination UDP port used to outgoing messages matches the source port of the received messages. This leads to the commissioner's ACK messages not being processed, causing the BR side to repeatedly retransmit the same udp_rx message.

To resolve this issue, we use the default MM port as the only supported destination port for UDP_RX messages

### DIFF
--- a/include/commissioner/defines.hpp
+++ b/include/commissioner/defines.hpp
@@ -124,6 +124,11 @@ static constexpr uint16_t kDefaultAeUdpPort = 1001;
 static constexpr uint16_t kDefaultNmkpUdpPort = 1002;
 
 /**
+ * The default management UDP Port used by TMF transaction Thread devices if not specified.
+ */
+static constexpr uint16_t kDefaultMmPort = 61631;
+
+/**
  * If using radio 915Mhz. Default radio freq of Thread is 2.4Ghz.
  *
  * Used to encoding ChannelMask TLV.
--- a/include/commissioner/defines.hpp
+++ b/include/commissioner/defines.hpp
@@ -124,7 +124,7 @@ static constexpr uint16_t kDefaultAeUdpPort = 1001;
 static constexpr uint16_t kDefaultNmkpUdpPort = 1002;
 
 /**
- * The default management UDP Port used by TMF transaction Thread devices if not specified.
+ * The default management UDP Port used by TMF transaction in Thread devices.
  */
 static constexpr uint16_t kDefaultMmPort = 61631;
 
--- a/src/library/udp_proxy.cpp
+++ b/src/library/udp_proxy.cpp
@@ -73,6 +73,7 @@ Error ProxyEndpoint::Send(const ByteArray &aRequest, MessageSubType aSubType)
     Error         error;
     coap::Request udpTx{coap::Type::kNonConfirmable, coap::Code::kPost};
     ByteArray     udpPayload;
+    uint16_t      sockPort;
 
     (void)aSubType;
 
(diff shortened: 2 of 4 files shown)


src/library/udp_proxy.cpp
```diff
@@ -81,7 +82,8 @@ Error ProxyEndpoint::Send(const ByteArray &aRequest, MessageSubType aSubType)
 
     VerifyOrExit(mBrClient.IsConnected(), error = ERROR_INVALID_STATE("not connected to the border agent"));
 
-    utils::Encode<uint16_t>(udpPayload, mBrClient.GetDtlsSession().GetLocalPort());
+    sockPort = (mSockPort == 0) ? mBrClient.GetDtlsSession().GetLocalPort() : GetSockPort();
```
wgtdkp, line 85: The `mSockPort` will be always not `0` once it received a `*.ans` from an end device, because there is no code to set `mSockPort` back to `0` explicitly.

It's probably just simpler to use the const int `kDefaultMmPort` for now? Since this is the only port the commissioner should be bind to.

src/library/udp_proxy.cpp
```diff
@@ -165,6 +167,7 @@ void ProxyClient::HandleUdpRx(const coap::Request &aUdpRx)
 
     mEndpoint.SetPeerAddr(peerAddr);
     mEndpoint.SetPeerPort(peerPort);
+    mEndpoint.SetSockPort(peerPort);
```
wgtdkp, line 170: Check that the `udpEncap` message is always setting the destination port to `kDefaultMmPort`? This is the simplest fix for now:

```
peerPort = utils::Decode<uint16_t>(udpEncap->GetValue());
destPort = utils::Decode<uint16_t>(&udpEncap->GetValue()[2], 2);
VerifyOrExit(destPort == kDefaultMmPort, ERROR_UNIMPLEMENTED("dropping UDP_RX.ntf message to port {}: support only port {}", destPort, kDefaultMmPort));
```

wgtdkp (commented):
Thanks @ZhangLe2016 The fix looks good to me overall. But see my comments.

A more proper fix is probably let the Coap::Request carry the destination port along with it, but it will require larger change.

ZhangLe2016:
Thanks @wgtdkp's reviewing and suggestions for the short term resolution of this problem. based on your comments, I have updated the PR.

wgtdkp (approved):
LGTM :+1:

include/commissioner/defines.hpp
```diff
@@ -123,6 +123,11 @@ static constexpr uint16_t kDefaultAeUdpPort = 1001;
  */
 static constexpr uint16_t kDefaultNmkpUdpPort = 1002;
 
+/**
+ * The default management UDP Port used by TMF transaction Thread devices if not specified.
```
bukepo, line 127: Do we allow specifying the port? It looks like it's hard coded in the implementation.

wgtdkp: Currently no, my plan is to support all/any port which the request is targeting, but it would require more changes to the coap library. Let this be a short-term workaround to unblock the diag command development. Thoughts?

ZhangLe2016: In the OT implementation, the sockport is also specified to default management port.

https://github.com/openthread/openthread/blob/main/src/core/thread/tmf.hpp#L91

bukepo: My concern is on the comment. It looks like this port can be specified.

ZhangLe2016: Updated Thanks @bukepo's comment.

bukepo (approved):

jwhui:
> A more proper fix is probably let the Coap::Request carry the destination port along with it, but it will require larger change.

I agree with @wgtdkp 's comment here, but I'm fine with the short term fix.

jwhui (approved):

PR #280, opened 2024-05-27
[cli] add DIAG_GET reset command into cli

ZhangLe2016:
The DIAG_GET reset command can reset the diagnostic TLVs:

* MAC Counters (9)
* MLE Counters (34)

> diag reset --flag 512
diag reset --flag 512
[done]

### DIFF
--- a/src/app/cli/interpreter.cpp
+++ b/src/app/cli/interpreter.cpp
@@ -200,6 +200,7 @@ const std::map<std::string, Interpreter::Evaluator> &Interpreter::mEvaluatorMap 
     {"announce", &Interpreter::ProcessAnnounce},   {"panid", &Interpreter::ProcessPanId},
     {"energy", &Interpreter::ProcessEnergy},       {"exit", &Interpreter::ProcessExit},
     {"quit", &Interpreter::ProcessExit},           {"help", &Interpreter::ProcessHelp},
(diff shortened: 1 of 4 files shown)


src/app/cli/interpreter.cpp
```diff
@@ -277,6 +278,7 @@ const std::map<std::string, std::string> &Interpreter::mUsageMap = *new std::map
               "panid conflict <panid>"},
     {"energy", "energy scan <channel-mask> <count> <period> <scan-duration> <dst-addr>\n"
                "energy report [<dst-addr>]"},
+    {"diag", "diag reset [--rloc <rloc>] [--flag <tlvs-flag>]"},
```
wgtdkp, line 281:
```suggestion
    {"diag", "diag reset <rloc> <tlvs-flag>"},
```

All the other commands don't use the options like "--rloc" or "--flag". Do we really need those here?

ZhangLe2016: The main purpose of adding options is to allow users to omit specifying the order of parameters. In such cases, users could not set the value of option `rloc`, and the leader’s `aloc` will be used by default.

ZhangLe2016:
This PR is too old, let me close it and resubmit a new one.

PR #281, opened 2024-05-27
[api] enable a DIAG_GET reset command into commissioner

ZhangLe2016:
the DIAG_GET reset command is a key functionality of diagnositic feature in TMF, it can be used to reset MAC counters (9) and MLE counters (34).

User can use reset/request commands to do the link stats on MAC and Mesh layer.


ZhangLe2016:
let me close this one and resubmit a new one to replace it. Thanks.

PR #283, opened 2024-06-19
[api] add a new api for getting the mesh local prefix

ZhangLe2016:
The primary purpose of providing this API is to enable external apps to generate the mesh local address(pefix + rloc) of a Thread device, which serves as the destination address for external commissioner to send UDP proxy commands.

Tested with CLI command:
> config set pskc 445f2b5ca6f2a93a55ce570a70efeecb
[done]
> start 192.168.9.2 49154
[done]
> meshlocalprefix
fd2b471153fcbe46
[done]

### DIFF
--- a/include/commissioner/commissioner.hpp
+++ b/include/commissioner/commissioner.hpp
@@ -1146,6 +1146,22 @@ public:
      *         git repository.
      */
     static std::string GetVersion(void);
+
+    /**
+     * @brief Asynchronously request mesh local prefix of the Thread Network.
+     *
+     * @param[in, out] aHandler  A handler of all errors; Guaranteed to be called.
+     */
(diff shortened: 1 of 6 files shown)


wgtdkp:
The mesh-local prefix can already be retrieved with the `GetActiveDataset()` API, right?

ZhangLe2016:
> The mesh-local prefix can already be retrieved with the `GetActiveDataset()` API, right?

do you mean there is no need to provide a dedicated API for the mesh-local prefix?

ZhangLe2016:
Discussed with @wgtdkp, the mesh local prefix could be obtained by `GetActiveDataset()` from the app layer, so lets close this PR.

PR #286, opened 2024-07-04
[coap] handle zero coap token length

wgtdkp:
There could be potential out of bound error if the token length in CoAP header is zero (`&aBuf[offset]` could be referring to the next element after the end of the buffer).

Note that the added test can pass without the fix in coap.cpp but it's probably because the toolchain has some optimization to avoid such issues...


superwhd (approved):
LGTM 👍

jwhui (approved):

PR #287, opened 2024-07-15
[swig] fix the SWIG ignores for java

wgtdkp:
The commissioner.i file is behind the APIs in include/commissioner/.

Also ignores `PrintTo` given it's only for googletest.


ZhangLe2016 (approved):
LGTM

jwhui (approved):

PR #303, opened 2024-10-16
[diag] add network diagnostic data definition and parsing

ZhangLe2016:
supported TLVs:
- Ipv6addressList
- ChildIpv6AddressList
- Mode
- ChildTable
- LeaderData
- Route64
- MacCounters
- Eui64
- ExtMacAddress
- MacAddress

### DIFF
--- a/include/commissioner/network_diag_data.hpp
+++ b/include/commissioner/network_diag_data.hpp
@@ -76,14 +76,6 @@ struct ChildTableEntry
 };
 
 /**
- * @brief IPv6 Addresses List
- */
-struct Ipv6AddressList
-{
-    std::vector<std::string> mIpv6Addresses;
-};
-
-/**
  * @brief Leader Data
  */
 struct LeaderData
--- a/include/commissioner/network_diag_data.hpp
+++ b/include/commissioner/network_diag_data.hpp
@@ -47,13 +47,6 @@ namespace ot {
 
 namespace commissioner {
 
-static constexpr uint8_t kChildTableEntryBytes = 3;
-static constexpr uint8_t kIpv6AddressBytes     = 16;
-static constexpr uint8_t kLeaderDataBytes      = 8;
-static constexpr uint8_t kMacCountersBytes     = 36;
-static constexpr uint8_t kRloc16Bytes          = 2;
-static constexpr uint8_t kRouterIdMaskBytes    = 8;
-
 /**
  * @brief Mode Data
  */
@@ -142,22 +135,21 @@ struct MacCounters
  */
 struct NetDiagData
 {
+    ModeData                          mMode;
+    ByteArray                         mEui64;
     ByteArray                         mExtMacAddress;
     uint16_t                          mMacAddress = 0;
-    ModeData                          mMode;
     Route64                           mRoute64;
     LeaderData                        mLeaderData;
+    MacCounters                       mMacCounters;
     std::vector<std::string>          mAddrs;
     std::vector<ChildTableEntry>      mChildTable;
-    ByteArray                         mEui64;
-    ByteArray                         mTlvTypeList;
     std::vector<ChildIpv6AddressInfo> mChildIpv6AddrsInfoList;
-    MacCounters                       mMacCounters;
 
     /**
-     * Indicates which fields are included in the dataset.
+     * Indicates which fields are included in the object.
      */
-    uint64_t mPresentFlags;
+    uint64_t mPresentFlags = 0;
 
     static constexpr uint64_t kExtMacAddressBit    = (1ull << 0);
     static constexpr uint64_t kMacAddressBit       = (1ull << 1);
@@ -167,13 +159,12 @@ struct NetDiagData
     static constexpr uint64_t kIpv6AddressBit      = (1ull << 5);
     static constexpr uint64_t kChildTableBit       = (1ull << 6);
     static constexpr uint64_t kEui64Bit            = (1ull << 7);
-    static constexpr uint64_t kTlvTypeBit          = (1ull << 8);
-    static constexpr uint64_t kMacCountersBit      = (1ull << 9);
-    static constexpr uint64_t kChildIpv6AddressBit = (1ull << 10);
+    static constexpr uint64_t kMacCountersBit      = (1ull << 8);
+    static constexpr uint64_t kChildIpv6AddressBit = (1ull << 9);
 };
 
 } // namespace commissioner
 
 } // namespace ot
 
-#endif // OT_COMM_NETWORK_DIAG_TLVS_HPP_
+#endif // OT_COMM_NETWORK_DIAG_DATA_HPP_
--- a/include/commissioner/network_diag_data.hpp
+++ b/include/commissioner/network_diag_data.hpp
@@ -107,7 +107,7 @@ struct Route64
 struct ChildIpv6AddressInfo
 {
     uint16_t                 mRloc16 = 0;
-    std::vector<std::string> mAddrs;
+    std::vector<std::string> mAddresses;
 };
 
 /**
@@ -142,7 +142,7 @@ struct NetDiagData
     Route64                           mRoute64;
     LeaderData                        mLeaderData;
     MacCounters                       mMacCounters;
-    std::vector<std::string>          mAddrs;
+    std::vector<std::string>          mAddresses;
     std::vector<ChildTableEntry>      mChildTable;
     std::vector<ChildIpv6AddressInfo> mChildIpv6AddrsInfoList;
 
--- a/include/commissioner/network_diag_data.hpp
+++ b/include/commissioner/network_diag_data.hpp
@@ -31,8 +31,8 @@
  *   This file defines the types of Thread Network Diagnostic TLVs used for network diagnostics.
  */
 
-#ifndef OT_COMM_NETWORK_DIAG_TLVS_HPP_
-#define OT_COMM_NETWORK_DIAG_TLVS_HPP_
+#ifndef OT_COMM_NETWORK_DIAG_DATA_HPP_
+#define OT_COMM_NETWORK_DIAG_DATA_HPP_
 
 #include <cstddef>
 #include <cstdint>
--- a/include/commissioner/network_diag_data.hpp
+++ b/include/commissioner/network_diag_data.hpp
@@ -104,7 +104,7 @@ struct Route64
 /**
  * @brief Child IPv6 Address Info
  */
-struct ChildIpv6AddressInfo
+struct ChildIpv6AddrInfo
 {
     uint16_t                 mRloc16 = 0;
     std::vector<std::string> mAddresses;
@@ -135,32 +135,32 @@ struct MacCounters
  */
 struct NetDiagData
 {
-    ModeData                          mMode;
-    ByteArray                         mEui64;
-    ByteArray                         mExtMacAddress;
-    uint16_t                          mMacAddress = 0;
-    Route64                           mRoute64;
-    LeaderData                        mLeaderData;
-    MacCounters                       mMacCounters;
-    std::vector<std::string>          mAddresses;
-    std::vector<ChildTableEntry>      mChildTable;
-    std::vector<ChildIpv6AddressInfo> mChildIpv6AddrsInfoList;
+    ModeData                       mMode;
+    ByteArray                      mEui64;
+    ByteArray                      mExtMacAddr;
+    uint16_t                       mMacAddr = 0;
+    Route64                        mRoute64;
+    LeaderData                     mLeaderData;
+    MacCounters                    mMacCounters;
+    std::vector<std::string>       mAddrs;
+    std::vector<ChildTableEntry>   mChildTable;
+    std::vector<ChildIpv6AddrInfo> mChildIpv6AddrsInfoList;
 
     /**
      * Indicates which fields are included in the object.
      */
     uint64_t mPresentFlags = 0;
 
-    static constexpr uint64_t kExtMacAddressBit    = (1ull << 0);
-    static constexpr uint64_t kMacAddressBit       = (1ull << 1);
-    static constexpr uint64_t kModeBit             = (1ull << 2);
-    static constexpr uint64_t kRoute64Bit          = (1ull << 3);
-    static constexpr uint64_t kLeaderDataBit       = (1ull << 4);
-    static constexpr uint64_t kIpv6AddressBit      = (1ull << 5);
-    static constexpr uint64_t kChildTableBit       = (1ull << 6);
-    static constexpr uint64_t kEui64Bit            = (1ull << 7);
-    static constexpr uint64_t kMacCountersBit      = (1ull << 8);
-    static constexpr uint64_t kChildIpv6AddressBit = (1ull << 9);
+    static constexpr uint64_t kExtMacAddrBit             = (1ull << 0);
+    static constexpr uint64_t kMacAddrBit                = (1ull << 1);
+    static constexpr uint64_t kModeBit                   = (1ull << 2);
+    static constexpr uint64_t kRoute64Bit                = (1ull << 3);
+    static constexpr uint64_t kLeaderDataBit             = (1ull << 4);
+    static constexpr uint64_t kAddrsBit                  = (1ull << 5);
+    static constexpr uint64_t kChildTableBit             = (1ull << 6);
+    static constexpr uint64_t kEui64Bit                  = (1ull << 7);
+    static constexpr uint64_t kMacCountersBit            = (1ull << 8);
+    static constexpr uint64_t kChildIpv6AddrsInfoListBit = (1ull << 9);
 };
 
 } // namespace commissioner
--- a/include/commissioner/network_diag_data.hpp
+++ b/include/commissioner/network_diag_data.hpp
@@ -43,16 +43,16 @@
 #include "defines.hpp"
 #include "error.hpp"
 
-#define CHILD_TABLE_ENTRY_BYTES 3
-#define IPV6_ADDRESS_BYTES 16
-#define LEADER_DATA_BYTES 8
-#define MAC_COUNTERS_BYTES 36
-#define RLOC16_BYTES 2
-#define ROUTER_ID_MASK_BYTES 8
-
 namespace ot {
 
 namespace commissioner {
+
+static constexpr uint8_t kChildTableEntryBytes = 3;
+static constexpr uint8_t kIpv6AddressBytes     = 16;
+static constexpr uint8_t kLeaderDataBytes      = 8;
+static constexpr uint8_t kMacCountersBytes     = 36;
+static constexpr uint8_t kRloc16Bytes          = 2;
+static constexpr uint8_t kRouterIdMaskBytes    = 8;
 
 /**
  * @brief Mode Data
@@ -117,12 +117,12 @@ struct Route64
 };
 
 /**
- * @brief Child IPv6 Address List
+ * @brief Child IPv6 Address Info
  */
-struct ChildIpv6AddressList
+struct ChildIpv6AddressInfo
 {
-    uint16_t        mRloc16 = 0;
-    Ipv6AddressList mIpv6AddressList;
+    uint16_t                 mRloc16 = 0;
+    std::vector<std::string> mAddrs;
 };
 
 /**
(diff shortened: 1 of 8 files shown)


include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,187 @@
...
+    ByteArray                    mExtMacAddress;
+    uint16_t                     mMacAddress = 0;
+    ModeData                     mMode;
+    Route64                      mRoute64;
+    LeaderData                   mLeaderData;
+    Ipv6AddressList              mIpv6AddressList;
+    std::vector<ChildTableEntry> mChildTable;
+    ByteArray                    mEui64;
+    ByteArray                    mTlvTypeList;
+    ChildIpv6AddressList         mChildIpv6AddressList;
```
wgtdkp, line 162:
```suggestion
    int mChildRloc16;
    std::vector<std::string>         mChildAddrs;
```

And remove the `ChildIpv6AddressList` class

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,187 @@
...
+ *    POSSIBILITY OF SUCH DAMAGE.
+ */
+
+/**
+ * @file
+ *   This file defines the types of Thread Network Diagnostic TLVs used for network diagnostics.
+ */
+
+#ifndef OT_COMM_NETWORK_DIAG_TLVS_HPP_
+#define OT_COMM_NETWORK_DIAG_TLVS_HPP_
```
wgtdkp, line 35:
```suggestion
#ifndef OT_COMM_NETWORK_DIAG_DATA_HPP_
#define OT_COMM_NETWORK_DIAG_DATA_HPP_
```

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,187 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <stdbool.h>
+#include <string>
+#include <vector>
+
+#include "defines.hpp"
+#include "error.hpp"
+
+#define CHILD_TABLE_ENTRY_BYTES 3
```
wgtdkp, line 46: We prefer `static constexpr int` over `#define`
```suggestion
static constexpr int CHILD_TABLE_ENTRY_BYTES = 3;
```

Please update for all constants and move them into the ot::commissioner:: namespace

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,187 @@
...
+ * `mPresentFlags`.
+ */
+struct NetDiagData
+{
+    ByteArray                    mExtMacAddress;
+    uint16_t                     mMacAddress = 0;
+    ModeData                     mMode;
+    Route64                      mRoute64;
+    LeaderData                   mLeaderData;
+    Ipv6AddressList              mIpv6AddressList;
```
wgtdkp, line 158:
```suggestion
    std::vector<std::string>              mAddrs;
```

include/commissioner/network_diag_data.hpp
ZhangLe2016: no, we cannot separate the `ChildIpv6AddressList` to `mChildRloc16` and `mChildAddrs` here. Because a Thread device may have more than one child, your change cannot support it.

wgtdkp: I see why you want the `ChildIpv6AddressList` now, you want to use a single present flag for a single member field, right? I think it's okay for this purpose. But still you don't need the `Ipv6AddressList` class, pleae delete that one.

Besides, I think the name "ChildAddressInfo" is better than ChildIpv6AddressList as it includes the RLOC integer.

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,187 @@
...
+    std::vector<RouteDataEntry> mRouteData;
+};
+
+/**
+ * @brief Child IPv6 Address List
+ */
+struct ChildIpv6AddressList
+{
+    uint16_t        mRloc16 = 0;
+    Ipv6AddressList mIpv6AddressList;
```
wgtdkp, line 125: How can we identify which Child has those address? Should there be a `Child ID` or something similar?

include/commissioner/network_diag_data.hpp
ZhangLe2016: done

include/commissioner/network_diag_data.hpp
ZhangLe2016: done

include/commissioner/network_diag_data.hpp
ZhangLe2016: we can identify the child with rloc16

include/commissioner/network_diag_data.hpp
ZhangLe2016: I think the previous definition of the `ChildIpv6AddressList ` is incorrect.
it should be defined as follow:
```
std::vector<ChildIpv6AddressInfo> mChildIpv6AddrsInfoList;
```
`ChildIpv6AddressInfo` represents the address information recorded by a child of the device, including its RLOC (Routing Locator) and addresses.  std::vector<ChildIpv6AddressInfo> records the address information of all the children of the device.

include/commissioner/network_diag_data.hpp
wgtdkp: Them how can we associate those addresses with a ChildTableEntry? There doesn't seem to be a rloc16 in ChildTableEntry...

ZhangLe2016: Although we can use `rloc16 & 0x1FF` to generate the `child ID` which can match with `ChildTableEntry `'s address, in practical applications, we still need to use RLOC16 to identify nodes in the network. In other words, the child ID in the Child Table Entry and the router ID of the device itself are used to generate the RLOC16 of that child.

src/library/commissioner_impl.cpp
```diff
@@ -1955,6 +1955,288 @@ ByteArray CommissionerImpl::GetCommissionerDatasetTlvs(uint16_t aDatasetFlags)
...
+
+    while (offset < length)
+    {
+        ChildTableEntry entry;
+        VerifyOrExit(offset + kChildTableEntryBytes <= length,
+                     error = ERROR_BAD_FORMAT("premature end of Child Table"));
+        entry.mTimeout             = 1 << (((aBuf[offset] & 0xF8) >> 3) - 4);
+        entry.mIncomingLinkQuality = (aBuf[offset] & 0x06) >> 1;
+        entry.mChildId             = ((aBuf[offset] & 0x01) << 9) | aBuf[offset + 1];
+        SuccessOrExit(error = DecodeModeData(entry.mModeData, {aBuf.begin() + offset + 2, aBuf.begin() + offset + 3}));
```
wgtdkp, line 2152: Can you document the TLV value format here? It's hard to follow the code.

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,179 @@
...
+
+#include "defines.hpp"
+#include "error.hpp"
+
+namespace ot {
+
+namespace commissioner {
+
+static constexpr uint8_t kChildTableEntryBytes = 3;
+static constexpr uint8_t kIpv6AddressBytes     = 16;
```
wgtdkp, line 51: Is this necessary being part of the library API? Will an client app need to use this value? If no, please move this constant to the class where it is used.

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,179 @@
...
+#include <vector>
+
+#include "defines.hpp"
+#include "error.hpp"
+
+namespace ot {
+
+namespace commissioner {
+
+static constexpr uint8_t kChildTableEntryBytes = 3;
```
wgtdkp, line 50: Ditto, is this useful for client apps?

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,179 @@
...
+static constexpr uint8_t kIpv6AddressBytes     = 16;
+static constexpr uint8_t kLeaderDataBytes      = 8;
```
wgtdkp, line 52: Ditto, this doesn't seem useful for client apps.

wgtdkp (commented):
Is it possible to add unit tests in commissioner_impl_test.cpp?

src/library/commissioner_impl.cpp
ZhangLe2016: done

src/library/commissioner_impl.cpp
```diff
@@ -1955,6 +1955,288 @@ ByteArray CommissionerImpl::GetCommissionerDatasetTlvs(uint16_t aDatasetFlags)
...
+    Error                error;
+    size_t               length = aBuf.size();
+    ChildIpv6AddressInfo childIpv6AddrsInfo;
+
+    VerifyOrExit((length - kRloc16Bytes) % kIpv6AddressBytes == 0,
+                 error = ERROR_BAD_FORMAT("premature end of Child IPv6 Address"));
+    childIpv6AddrsInfo.mRloc16 = utils::Decode<uint16_t>(aBuf.data(), kRloc16Bytes);
+
+    SuccessOrExit(error = DecodeIpv6AddressList(childIpv6AddrsInfo.mAddrs, {aBuf.begin() + kRloc16Bytes, aBuf.end()}));
+    aChildIpv6AddressInfoList.emplace_back(childIpv6AddrsInfo);
```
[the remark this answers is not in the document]
ZhangLe2016: no, the `DecodeChildIpv6AddressList ` may be call multiple times, and the reference value of `std::vector<ChildIpv6AddressInfo>` will be added new childIpv6AddrsInfo object at its end.

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,179 @@
...
+    uint16_t                          mMacAddress = 0;
+    ModeData                          mMode;
+    Route64                           mRoute64;
+    LeaderData                        mLeaderData;
+    std::vector<std::string>          mAddrs;
+    std::vector<ChildTableEntry>      mChildTable;
+    ByteArray                         mEui64;
+    ByteArray                         mTlvTypeList;
+    std::vector<ChildIpv6AddressInfo> mChildIpv6AddrsInfoList;
+    MacCounters                       mMacCounters;
```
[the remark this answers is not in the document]
ZhangLe2016: done.

src/library/commissioner_impl.cpp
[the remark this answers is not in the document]
wgtdkp: Could you please point me to the multiple call sites of `DecodeChildIpv6AddressList`?

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,179 @@
...
+{
+    ByteArray                         mExtMacAddress;
+    uint16_t                          mMacAddress = 0;
+    ModeData                          mMode;
+    Route64                           mRoute64;
+    LeaderData                        mLeaderData;
+    std::vector<std::string>          mAddrs;
+    std::vector<ChildTableEntry>      mChildTable;
+    ByteArray                         mEui64;
+    ByteArray                         mTlvTypeList;
```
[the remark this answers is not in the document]
ZhangLe2016: the TLV named `TlvTypeList` is used for collecting the request/query TLVs' type, we just use it before executing request and query commands, so I think we can remove it from here.

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,179 @@
...
+    std::vector<ChildIpv6AddressInfo> mChildIpv6AddrsInfoList;
+    MacCounters                       mMacCounters;
+
+    /**
+     * Indicates which fields are included in the dataset.
```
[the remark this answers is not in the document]
ZhangLe2016: done.

src/library/commissioner_impl.cpp
```diff
@@ -1955,6 +1955,288 @@ ByteArray CommissionerImpl::GetCommissionerDatasetTlvs(uint16_t aDatasetFlags)
...
+    ByteArray tlvTypes;
+
+    if (aDiagTlvFlags & NetDiagData::kExtMacAddressBit)
+    {
+        EncodeTlvType(tlvTypes, tlv::Type::kNetworkDiagExtMacAddress);
+    }
+    if (aDiagTlvFlags & NetDiagData::kMacAddressBit)
+    {
+        EncodeTlvType(tlvTypes, tlv::Type::kNetworkDiagMacAddress);
+    }
```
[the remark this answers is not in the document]
ZhangLe2016: done.  Thanks.

include/commissioner/network_diag_data.hpp
ZhangLe2016: I think these const variables only be used by `commissioner_impl`, i have removed them to  `commissioner_impl`. Thanks.

include/commissioner/network_diag_data.hpp
ZhangLe2016: done.

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,179 @@
...
+namespace ot {
+
+namespace commissioner {
+
+static constexpr uint8_t kChildTableEntryBytes = 3;
+static constexpr uint8_t kIpv6AddressBytes     = 16;
+static constexpr uint8_t kLeaderDataBytes      = 8;
+static constexpr uint8_t kMacCountersBytes     = 36;
+static constexpr uint8_t kRloc16Bytes          = 2;
+static constexpr uint8_t kRouterIdMaskBytes    = 8;
```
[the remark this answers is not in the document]
ZhangLe2016: done.

src/library/commissioner_impl.cpp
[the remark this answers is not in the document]
ZhangLe2016: oh, I mistakenly thought that `tlvSet` in line 2029 could hold a map<type, List<Tlv>>, but it actually can only hold a map<type, Tlvptr>. This means I need to make some changes. Thanks for pointing that out!

include/commissioner/network_diag_data.hpp
```diff
@@ -0,0 +1,170 @@
...
+struct NetDiagData
+{
+    ModeData                          mMode;
+    ByteArray                         mEui64;
+    ByteArray                         mExtMacAddress;
+    uint16_t                          mMacAddress = 0;
+    Route64                           mRoute64;
+    LeaderData                        mLeaderData;
+    MacCounters                       mMacCounters;
+    std::vector<std::string>          mAddrs;
```
wgtdkp, line 145: Some place we use "Addr" while some place we use "Address", it doesn't matter which one is used, but it would be better to make it consistent. Looking at existing code, "Addr" is always used for addresses. Can we update the code in PR to follow the existing naming? Thanks

src/library/commissioner_impl.cpp
```diff
@@ -1955,6 +1962,314 @@ ByteArray CommissionerImpl::GetCommissionerDatasetTlvs(uint16_t aDatasetFlags)
...
+    return tlvTypes;
+}
+
+Error CommissionerImpl::DecodeNetDiagData(NetDiagData &aNetDiagData, const ByteArray &aPayload)
+{
+    Error       error;
+    tlv::TlvSet tlvSet;
+    NetDiagData diagData;
+    // Clear all data fields
+    diagData.mPresentFlags = 0;
```
wgtdkp, line 2028: nit: this can be removed given the default value for `mPresentFlags` is already `0`

include/commissioner/network_diag_data.hpp
wgtdkp: This means the client app developers need to know that they need to get the child id from `rloc16 & 0x1FF`, this is a knowledge barrier. Can we include also the `ChildId` in the `ChildIpv6AddressInfo` class? Just as there is a `ChildId` in `ChildTableEntry`. Or provide a member method `getChildId() const {...}`

wgtdkp (commented):
Thanks for resolving the comments!
In internal code review, we asked to mark comments resolved when you followed the suggestion in it and leave the other open. This is useful for reviewers to re-review the code and skip comments which has been resolved. It makes the review much easier.

Can you follow that practice in GitHub as well? For example, you can and should only close comments which you replied "done" :)

include/commissioner/network_diag_data.hpp
ZhangLe2016: Added a new field named `ChildId` into the `ChildTableEntry`.

include/commissioner/network_diag_data.hpp
ZhangLe2016: I see, I changed `addrs` to `address`

src/library/commissioner_impl.cpp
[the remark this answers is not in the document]
ZhangLe2016: I added a new function named `GetTlvListByType` into Tlv.cpp module which can fetch the TLV list like `ChildIpv6AddressList` into a Tlv List, and I used this method to solve this problem.

include/commissioner/network_diag_data.hpp
wgtdkp: No, if you look at include/commissioner/commissioner.hpp, "Addr" is used but not "Address"

ZhangLe2016:
> Is it possible to add unit tests in commissioner_impl_test.cpp?

Looks like the unit test cases need to call the API functions defined into `commissioner.hpp`. For this PR, the new functions are defined in private, I am not sure if we can add corresponding test cases for them, do you have any suggestions?

ZhangLe2016:
> > Is it possible to add unit tests in commissioner_impl_test.cpp?
> 
> Looks like the unit test cases need to call the API functions defined into `commissioner.hpp`. For this PR, the new functions are defined in private, I am not sure if we can add corresponding test cases for them, do you have any suggestions?

Added a new unit test case into `commissioner_impl_test.cpp`

src/library/tlv.hpp
```diff
@@ -217,7 +217,9 @@ class Tlv
 
 Error  GetTlvSet(TlvSet &aTlvSet, const ByteArray &aBuf, Scope aScope = Scope::kMeshCoP);
 TlvPtr GetTlv(tlv::Type aTlvType, const ByteArray &aBuf, Scope aScope = Scope::kMeshCoP);
-bool   IsDatasetParameter(bool aIsActiveDataset, tlv::Type aTlvType);
+Error  GetTlvListByType(TlvList &aTlvList, const ByteArray &aBuf, tlv::Type aTlvType, Scope aScope = Scope::kMeshCoP);
+
+bool IsDatasetParameter(bool aIsActiveDataset, tlv::Type aTlvType);
```
wgtdkp, line 222: nit: is the newline intentional?
```suggestion
Error  GetTlvListByType(TlvList &aTlvList, const ByteArray &aBuf, tlv::Type aTlvType, Scope aScope = Scope::kMeshCoP);
bool   IsDatasetParameter(bool aIsActiveDataset, tlv::Type aTlvType);
```

src/library/commissioner_impl_test.cpp
```diff
@@ -143,6 +145,43 @@ TEST(CommissionerImpl, NotImplementedApis)
...
+    EXPECT_EQ(diagData.mMode.mIsMtd, false);
+    EXPECT_EQ(diagData.mRoute64.mRouteData.size(), 9);
+    EXPECT_EQ(diagData.mAddrs.size(), 4);
+    EXPECT_EQ(diagData.mAddrs[0], "fd92:38a3:395d:1:f904:3dfe:b7b3:edf3");
+    EXPECT_EQ(diagData.mChildTable.size(), 3);
+    EXPECT_EQ(diagData.mChildTable[0].mChildId, 24);
+    EXPECT_EQ(diagData.mLeaderData.mRouterId, 33);
+    EXPECT_EQ(diagData.mChildIpv6AddrsInfoList.size(), 2);
+    EXPECT_EQ(diagData.mChildIpv6AddrsInfoList[0].mRloc16, 28696);
+    EXPECT_EQ(diagData.mChildIpv6AddrsInfoList[0].mChildId, 24);
```
wgtdkp, line 182: nit:
1. assert the child address as well
2. assert the second child informations as well

wgtdkp (approved):
Looks good to me overall, just minor comments!

src/library/commissioner_impl_test.cpp
ZhangLe2016: done.

wgtdkp: Better assert rloc, address and childId for both two Children

ZhangLe2016: done.

ZhangLe2016:
@jwhui Could you please review this PR? Thanks~~

jwhui (approved):

PR #317, opened 2025-08-18
[fmtlib] upgrade to 11.2.0 and add the const qualifier to a formatter specialization

ZhangLe2016:
upgrade the fmtlib to 11.2.0


wgtdkp (approved):
Thanks!

